PR #13, opened 2021-06-11
8268349: Provide clear run-time warnings about Security Manager deprecation

wangweij:
More loudly and precise warning messages when a security manager is either enabled at startup or installed at runtime.

This is new PR for the `openjdk/jdk17` repo copied from https://github.com/openjdk/jdk/pull/4400. A new commit is added.

---------
### Progress

### Issue
 * [JDK-8268349](https://bugs.openjdk.java.net/browse/JDK-8268349): Provide clear run-time warnings about Security Manager deprecation ⚠️ Issue is not open.

### Reviewers
 * [Lance Andersen](https://openjdk.java.net/census#lancea) (@LanceAndersen - **Reviewer**) ⚠️ Review applies to cb732f99ce13875c3c9aff3c8dacf1298dbec19b
 * [Sean Mullan](https://openjdk.java.net/census#mullan) (@seanjmullan - **Reviewer**) ⚠️ Review applies to 2d5b1ba53c6fb5db8346896b3fa3a23318bda5c3
 * [Alan Bateman](https://openjdk.java.net/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to 2d5b1ba53c6fb5db8346896b3fa3a23318bda5c3

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/13/head:pull/13` \
`$ git checkout pull/13`

`$ git checkout pull/13` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/13/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 13`

`$ git pr show -t 13`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/13.diff">https://git.openjdk.java.net/jdk17/pull/13.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/System.java
+++ b/src/java.base/share/classes/java/lang/System.java
@@ -61,7 +61,6 @@ import java.util.Properties;
 import java.util.PropertyPermission;
 import java.util.ResourceBundle;
 import java.util.Set;
-import java.util.WeakHashMap;
 import java.util.function.Supplier;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Stream;
@@ -391,7 +390,7 @@ public final class System {
             // security manager not allowed
             if (sm != null) {
                 throw new UnsupportedOperationException(
-                    "Runtime configured to disallow security manager");
+                    "The Security Manager API is deprecated and will be removed in a future release");
             }
         }
     }
@@ -2213,8 +2212,8 @@ public final class System {
         if (needWarning) {
             System.err.printf("""
                     WARNING: A Security Manager has been enabled on the command line
-                    WARNING: java.lang.SecurityManager is deprecated and will be removed in a future release
-                    WARNING: -Djava.security.manager%s will have no effect when java.lang.SecurityManager is removed
+                    WARNING: The Security Manager API is deprecated and will be removed in a future release
+                    WARNING: -Djava.security.manager%s will have no effect when the API is removed
                     """, smProp.isEmpty() ? "" : ("=" + smProp));
         }
 
--- a/src/java.base/share/classes/java/lang/System.java
+++ b/src/java.base/share/classes/java/lang/System.java
@@ -381,16 +381,16 @@ public final class System {
             String signature = caller.getName() + " (" + codeSource(caller) + ")";
             originalErrStream.printf("""
                     WARNING: A terminally deprecated method in java.lang.System has been called
-                    WARNING: java.lang.System::setSecurityManager has been called by %s
+                    WARNING: System::setSecurityManager has been called by %s
                     WARNING: Please consider reporting this to the maintainers of %s
-                    WARNING: java.lang.System::setSecurityManager will be removed in a future release
+                    WARNING: System::setSecurityManager will be removed in a future release
                     """, signature, caller.getName());
             implSetSecurityManager(sm);
         } else {
             // security manager not allowed
             if (sm != null) {
                 throw new UnsupportedOperationException(
-                    "The Security Manager API is deprecated and will be removed in a future release");
+                    "The Security Manager is deprecated and will be removed in a future release");
             }
         }
     }
@@ -2210,11 +2210,9 @@ public final class System {
         }
 
         if (needWarning) {
-            System.err.printf("""
-                    WARNING: A Security Manager has been enabled on the command line
-                    WARNING: The Security Manager API is deprecated and will be removed in a future release
-                    WARNING: -Djava.security.manager%s will have no effect when the API is removed
-                    """, smProp.isEmpty() ? "" : ("=" + smProp));
+            System.err.println("""
+                    WARNING: A command line option has enabled the Security Manager
+                    WARNING: The Security Manager is deprecated and will be removed in a future release""");
         }
 
         originalErrStream = System.err;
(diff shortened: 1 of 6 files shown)


LanceAndersen (approved):

test/jdk/java/lang/System/SecurityManagerWarnings.java
```diff
@@ -31,6 +31,8 @@
 import jdk.test.lib.process.OutputAnalyzer;
 import jdk.test.lib.process.ProcessTools;
 
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
 import java.security.Permission;
 
 public class SecurityManagerWarnings {
```
seanjmullan, line 38: This should probably have 8268349 in the @bug line.

test/jdk/java/security/ProtectionDomain/RecursionDebug.java
```diff
@@ -87,7 +87,5 @@ public static void main(String[] args) throws Exception {
             throw new Exception
                 ("Test with custom non-bootclass SecurityManager failed");
         }
-
-        System.setSecurityManager(null);
```
seanjmullan, line 91: Why did this line need to be removed?

test/jdk/java/lang/System/SecurityManagerWarnings.java
wangweij: OK.

test/jdk/java/security/ProtectionDomain/RecursionDebug.java
wangweij: This is where the `setSecurityManager` method is called again when a security manager has already been installed, and the `getProtectionDomain` permission is needed. While I can add the new permission into the policy file, my understanding is that this line was just a clean-up and the test was about implementation of `ProtectionDomain::toString` (esp the `seeAllp` method called by it) when debug is on.

seanjmullan: Ok, thanks for the explanation. I think it should be ok to remove this line then.

seanjmullan (approved):

src/java.base/share/classes/java/lang/System.java
```diff
@@ -324,6 +326,16 @@ private static void checkIO() {
     private static native void setOut0(PrintStream out);
     private static native void setErr0(PrintStream err);
 
+    // Remember original System.err. setSecurityManager() warning goes here
+    private static volatile @Stable PrintStream originalErrStream = null;
```
AlanBateman, line 330: I'd probably use "initial" rather than "original" here but okay. You can drop setting it to null as that is its default value.

AlanBateman (approved):
Implementation changes/text is okay.

src/java.base/share/classes/java/lang/System.java
wangweij: Will do. Thanks.

wangweij:
A new commit. Much more exact checking of various warning messages.

src/java.base/share/classes/java/lang/System.java
```diff
@@ -362,16 +374,23 @@ private static void checkIO() {
      *       method.
      */
     @Deprecated(since="17", forRemoval=true)
+    @CallerSensitive
     public static void setSecurityManager(@SuppressWarnings("removal") SecurityManager sm) {
         if (allowSecurityManager()) {
-            System.err.println("WARNING: java.lang.System::setSecurityManager" +
-                    " is deprecated and will be removed in a future release.");
+            var caller = Reflection.getCallerClass();
+            String signature = caller.getName() + " (" + codeSource(caller) + ")";
```
jaikiran, line 381: Hello Weijun,

Given that the `codeSource()` method above is allowed to return `null`, there could be a case where the warning message printed would be something like:

> 
> WARNING: A terminally deprecated method in java.lang.System has been called
> WARNING: System::setSecurityManager has been called by foo.bar.Hello (null)
> WARNING: Please consider reporting this to the maintainers of foo.bar.Hello
> WARNING: System::setSecurityManager will be removed in a future release
> 

Would that be OK or do you think the presence of "(null)" be unnecessary and confusing? Maybe in such cases that line should just say "System::setSecurityManager has been called by foo.bar.Hello"?

Another minor nit - the variable is named `signature` which initially gave me an impression that it was the method signature of the caller code, but that isn't the case. Should this variable be renamed perhaps?

jaikiran:
Hello Sean, Weijung,

From what I have known, the Java/JDK code has always taken extra precaution when it comes to printing out potentially sensitive details like IP addresses and paths to file, like jar files in the log messages or exception stacktraces. In fact, one of the annoying things about some of the error messages that the JarFile API throws is that it doesn't even print out the jar file name, let alone the full path of the jar file which ran into issues. At least that was the case, unless that has changed in recent times. Furthermore, as you will surely know, to print out these details there's an security property which needs to be explicitly enabled ("jdk.includeInExceptions") with the right values.

Given all that, do you think that we should be printing the jar file paths in this WARNING message by default? I read the linked CSR, but I didn't see why the location of the jar or the name of the jar would be useful in this warning message. As long as the caller class (and perhaps the caller method) is printed, I think that should be enough of a summary on what's triggering this warning.

src/java.base/share/classes/java/lang/System.java
wangweij: Good catches. Will fix it. Thanks.

PR #18, opened 2021-06-11
8268591: a few runtime/jni tests don't need `/othervm`

iignatev:
Hi all,

could you please review this small and trivial patch that removes `/othervm` from 3 `runtime/jni` tests?
from JBS:
> `runtime/jni/registerNativesWarning/TestRegisterNativesWarning.java`, `checked/TestCheckedEnsureLocalCapacity.java`, and `atExit/TestAtExit.java` use `main/othervm` whereas in fact, they don't need `othervm`, and if we had CODETOOLS-7902658 implemented, they would have used `driver` mode. this RFE is to remove `/othervm`, so we at least won't waste time on creating a fresh JVM.

the patch explicitly sets `-Djava.library.path=` for the child processes so they are able to locate native libraries.

testing: `runtime/jni` on `{linux,windows,macosx}-x64`

Thanks,
-- Igor

---------
### Progress

### Issue
 * [JDK-8268591](https://bugs.openjdk.java.net/browse/JDK-8268591): a few runtime/jni tests don't need `/othervm`

### Reviewers
 * [Mikhailo Seledtsov](https://openjdk.java.net/census#mseledtsov) (@mseledts - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/18/head:pull/18` \
`$ git checkout pull/18`

`$ git checkout pull/18` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/18/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 18`

`$ git pr show -t 18`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/18.diff">https://git.openjdk.java.net/jdk17/pull/18.diff</a>

</details>

### DIFF
--- a/test/hotspot/jtreg/runtime/jni/atExit/TestAtExit.java
+++ b/test/hotspot/jtreg/runtime/jni/atExit/TestAtExit.java
@@ -32,7 +32,7 @@ import jdk.test.lib.process.ProcessTools;
  *          atexit handler fails as expected without crashing.
  *
  * @library /test/lib
- * @run main/othervm/native TestAtExit
+ * @run main/native TestAtExit
  */
 
 public class TestAtExit {
--- a/test/hotspot/jtreg/runtime/jni/atExit/TestAtExit.java
+++ b/test/hotspot/jtreg/runtime/jni/atExit/TestAtExit.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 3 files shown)


dholmes-ora:
Hi Igor,

The test classes, run by jtreg, load native libraries (though they could possibly be refactored so that is not necessary) so I would expect we would want othervm here because we do not want to load multiple unrelated libraries in the shared VM implied by "main".

I also don't understand why you now need to set java.library.path explicitly in some of the tests ??

Thanks,
David

iignatev:
> The test classes, run by jtreg, load native libraries (though they could possibly be refactored so that is not necessary) so I would expect we would want othervm here because we do not want to load multiple unrelated libraries in the shared VM implied by "main".

`jtreg` uses a separate classloader for each test, and native libraries are to be unloaded when CL is GC'ed, so there should be no unrelated libraries in shared VMs.

> I also don't understand why you now need to set java.library.path explicitly in some of the tests ??

frankly, I don't fully understand why it's needed, but these tests started to fail w/ `UnsatisfiedLinkError` when I removed `/othervm` and setting `java.library.path` fixed the problem. I assumed it's because we don't set/propagate `LD_LIBRARY_PATH` (or its equivalent), yet all my attempts to reproduce this behavior and the problem w/ a small test case failed, I'll dig deeper next week.

Thanks,
-- Igor

dholmes-ora (approved):
Thanks for the explanations Igor.

David

mseledts (approved):
Looks reasonable to me.

iignatev:
David, Misha, thank you for your reviewes!

/integrate

PR #25, opened 2021-06-11
8268602: a couple runtime/os tests don't check exit code

iignatev:
Hi all,

could you please review this trivial and test-only patch that adds checks of exit code to two `runtime/os` tests?
from JBS:
> `runtime/os/TestUseCpuAllocPath.java` and `AvailableProcessors.java` spawn new JVMs but don't check their exit code which might lead to both type-I and type-II errors

testing: ``runtime/os` on `{linux,windows,macosx}-x64`

Thanks,
-- Igor

---------
### Progress

### Issue
 * [JDK-8268602](https://bugs.openjdk.java.net/browse/JDK-8268602): a couple runtime/os tests don't check exit code

### Reviewers
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to abaabc29d955ff03c4f472f658b0faff8e352fff

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/25/head:pull/25` \
`$ git checkout pull/25`

`$ git checkout pull/25` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/25/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 25`

`$ git pr show -t 25`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/25.diff">https://git.openjdk.java.net/jdk17/pull/25.diff</a>

</details>

### DIFF
--- a/test/hotspot/jtreg/runtime/os/AvailableProcessors.java
+++ b/test/hotspot/jtreg/runtime/os/AvailableProcessors.java
@@ -88,6 +88,7 @@ public class AvailableProcessors {
                                    ProcessTools.getCommandLine(pb));
                 OutputAnalyzer output = ProcessTools.executeProcess(pb);
                 output.shouldContain(SUCCESS_STRING);
+                output.shouldHaveExitValue(0);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


dholmes-ora (approved):
The value of doing this seems marginal, but strictly speaking yes every test that launches  VM should check for a zero exit code. Arguably this should be built into the ProcessTools API so that we don't have to do it explicitly. The special case would be looking for a non-zero exit code.

In the case of failure it is more efficient to check the exit code first before potentially wasting time parsing output. But we probably don't care too much about that.

Thanks,
David

iignatev:
thanks, David. I've moved `OutputAnalyzer::shouldHaveExitValue` calls. I agree that it makes sense to update ProcessTools API to always check exit code; I'll file an RFE for that.

/integrate

PR #37, opened 2021-06-12
8268643: SVML lib shouldn't be generated when C2 is absent

DamonFool:
Hi all,

SVML lib shouldn't be generated when C2 is absent.
This is because it is only used by C2.

Thanks.
Best regards,
Jie

---------
### Progress

### Issue
 * [JDK-8268643](https://bugs.openjdk.java.net/browse/JDK-8268643): SVML lib shouldn't be generated when C2 is absent

### Reviewers
 * [Erik Joelsson](https://openjdk.java.net/census#erikj) (@erikj79 - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/37/head:pull/37` \
`$ git checkout pull/37`

`$ git checkout pull/37` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/37/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 37`

`$ git pr show -t 37`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/37.diff">https://git.openjdk.java.net/jdk17/pull/37.diff</a>

</details>


make/modules/jdk.incubator.vector/Lib.gmk
```diff
@@ -27,7 +27,7 @@ include LibCommon.gmk
 
 ################################################################################
 
-ifeq ($(call isTargetOs, linux windows)+$(call isTargetCpu, x86_64), true+true)
+ifeq ($(call isTargetOs, linux windows)+$(call isTargetCpu, x86_64)+$(INCLUDE_COMPILER2), true+true+true)
```
erikj79, line 30: Instead of introducing a new exported variable from spec.gmk, you can simply check for the JVM feature like this:

$(call check-jvm-feature, compiler2)

This macro will return "true" or "false" so can be used just like $(INCLUDE_COMPILER2) in the line above. Due to line length I would probably put it in a nested ifeq statement though.

make/autoconf/jvm-features.m4
```diff
@@ -565,6 +568,7 @@ AC_DEFUN_ONCE([JVM_FEATURES_SETUP],
   # missing any of them.
   ENABLE_CDS="true"
   INCLUDE_JVMCI="true"
+  INCLUDE_COMPILER2="true"
```
erikj79, line 571: I would recommend inverting this logic so that it works properly in the (unlikely) even that two or more JVM variants are built into the same JDK. I assume we want this library built if any JVM variant has c2.

DamonFool: Thanks @erikj79 for your review.

What do you mean by `two or more JVM variants are built into the same JDK` ?
Did you mean two different JVM variants can be built into just one JDK images?

As you suggested, I used to try to use `$(call check-jvm-feature, compiler2)` like this
```
diff --git a/make/modules/jdk.incubator.vector/Lib.gmk b/make/modules/jdk.incubator.vector/Lib.gmk
index 2626065..a176b9a 100644
--- a/make/modules/jdk.incubator.vector/Lib.gmk
+++ b/make/modules/jdk.incubator.vector/Lib.gmk
@@ -27,7 +27,7 @@ include LibCommon.gmk
 
 ################################################################################
 
-ifeq ($(call isTargetOs, linux windows)+$(call isTargetCpu, x86_64), true+true)
+ifeq ($(call isTargetOs, linux windows)+$(call isTargetCpu, x86_64)+$(call check-jvm-feature, compiler2), true+true+true)
   $(eval $(call SetupJdkLibrary, BUILD_LIBSVML, \
       NAME := svml, \
       CFLAGS := $(CFLAGS_JDKLIB), \
```
But it doesn't work.
Am I missing something?

Thanks.

DamonFool:
> Yes, there is still support for building multiple variants into the same
> image even though this is not something we currently do for any Oracle
> offering.

Sounds amazing.
Updated.
Thanks.

erikj79 (approved):

PR #39, opened 2021-06-14
8268293: VectorAPI cast operation on mask and shuffle is broken

jatin-bhateja:
Existing implementation of VectoMask.cast and VectorShuffle.cast operation do not generate correct mask compliant with the species passed as the argument.

This patch fixes the implementation and adds the corresponding cast operation test cases to VectorAPI regression suite.

---------
### Progress

### Issue
 * [JDK-8268293](https://bugs.openjdk.java.net/browse/JDK-8268293): VectorAPI cast operation on mask and shuffle is broken

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - **Reviewer**)
 * [Sandhya Viswanathan](https://openjdk.java.net/census#sviswanathan) (@sviswa7 - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/39/head:pull/39` \
`$ git checkout pull/39`

`$ git checkout pull/39` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/39/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 39`

`$ git pr show -t 39`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/39.diff">https://git.openjdk.java.net/jdk17/pull/39.diff</a>

</details>


jatin-bhateja:
Hi @PaulSandoz , Kindly let me know if there are review comments, I have withdrawn the PR from JDK-mainline.

PaulSandoz (commented):
This is looking much better.

The data providers don't need to provide a list of illegal and illegal as arguments to the test. Instead, you can have two data providers one providing the array of array of arguments for legal species, and the other for illegal. 

(A data provider returns Object[][] providing the array of arguments. The length of the outer array is the number of tests to run, so the test itself does not need to loop over the test cases.)

Then, have two corresponding tests, that gets supplied with a **single** src and target species like for the other conversion tests.

Then, tests can call a *kernel* that performs the looping, again copying the pattern of the existing tests.

It might be easier to first modify the data providers, then when that is working develop the kernel.

jatin-bhateja:
> This is looking much better.
> 
> The data providers don't need to provide a list of illegal and illegal as arguments to the test. Instead, you can have two data providers one providing the array of array of arguments for legal species, and the other for illegal.
> 
> (A data provider returns Object[][] providing the array of arguments. The length of the outer array is the number of tests to run, so the test itself does not need to loop over the test cases.)
> 
> Then, have two corresponding tests, that get supplied with a **single** src and target species like for the other conversion tests.
> 
> Then, tests can call a _kernel_ that performs the looping, again copying the pattern of the existing tests.
> 
> It might be easier to first modify the data providers, then when that is working develop the kernel.

Hi @PaulSandoz , thanks for your suggestions, I have split the data provider and tests, we now have separate +ve and -ve tests, each of which accepts a species and the array of legal/illegal species corresponding to it.

test/jdk/jdk/incubator/vector/AbstractVectorConversionTest.java
```diff
@@ -241,6 +241,29 @@ else if (src == double.class) {
...
+                            sps.add(dst);
+                        }
+                    } else {
+                        if (dst.length() != src.length()) {
+                            sps.add(dst);
+                        }
+                    }
+                }
+            }
+            args.add(new Object[]{src, sps});
```
PaulSandoz, line 263: Move the addition to `args` into the inner loop as:
```java
  args.add(new Object[]{src, dst});
```

Then, change the signature of the tests to be `(VectorSpecies src, VectorSpecies> dst)`.

Then, we can add invocation loops around the cast operation from `src` -> `dst` in the legal case to trigger compilation.

For the illegal case your test would reduce to:
```java
    @Test(dataProvider = "fixedShapeXSegmentedIllegalCastSpecies")
    static void maskCastNeg(VectorSpecies src, List<VectorSpecies> dst) {
        long val = (1L << (src.length() & 63)) - 1L;
        VectorMask mask = VectorMask.fromLong(src, val);
        try {
            mask.cast(sps);
            Assert.fail();
        } catch (IllegalArgumentException e) {
        }
    }
```

And in fact i believe you can push the body of this method into a method on `AbstractVectorConversionTest`, further reducing duplication, such as:
```java
    @Test(dataProvider = "fixedShapeXSegmentedIllegalCastSpecies")
    static void maskCastNeg(VectorSpecies src, List<VectorSpecies> dst) {
        maskCastFail(src, dest);
    }
```

Same technique can be applied to the non-exceptional test.

jatin-bhateja:
Hi @PaulSandoz , Thanks for the suggestions, patch has been updated accordingly. There is also another issue around limited intrinsification of mask casting operation as VectorSupport.convert is called only when BITSIZE of src/dst species exactly match which is an added constraint. I will be creating a separate RFE to fix that along with compiler side changes since we can generate optimal code for AVX512 mask casting operations.

test/jdk/jdk/incubator/vector/AbstractVectorConversionTest.java
```diff
@@ -511,4 +511,50 @@ static void copyConversionArray(Object src, int srcPos,
...
+    static <E,F> void legal_mask_cast_kernel(VectorSpecies<E> src, VectorSpecies<F> dst) {
+        for(int i = 0; i < INVOC_COUNT; i++) {
+            VectorMask<E> mask = VectorMask.fromLong(src, i);
+            VectorMask<F> res = mask.cast(dst);
+            Assert.assertEquals(res.toLong(), mask.toLong());
+        }
+    }
+
+    static <E,F> void illegal_mask_cast_kernel(VectorSpecies<E> src, VectorSpecies<F> dst) {
+        for(int i = 0; i < INVOC_COUNT; i++) {
```
PaulSandoz, line 524: We don't need the `INVOC_COUNT` loop for the illegal case, since the checks are performed in Java and not in the intrinsic.

test/jdk/jdk/incubator/vector/AbstractVectorConversionTest.java
```diff
@@ -245,22 +247,20 @@ else if (src == double.class) {
         List<Object[]> args = new ArrayList<>();
         for (Class<?> srcE : List.of(byte.class, short.class, int.class, long.class, float.class, double.class)) {
             VectorSpecies<?> src = VectorSpecies.of(srcE, srcShape);
-            List<VectorSpecies> sps = new ArrayList<VectorSpecies>();
             for (VectorShape dstShape : VectorShape.values()) {
                 for (Class<?> dstE : List.of(byte.class, short.class, int.class, long.class, float.class, double.class)) {
                     VectorSpecies<?> dst = VectorSpecies.of(dstE, dstShape);
                     if (legal) {
```
PaulSandoz, line 253: I think we just need to do `xnor` on `legal` and `dst.length() == src.length()`, we could simplify with:
```
// add if legal and equal lengths, or illegal and unequal lengths 
if (legal == (dst.length() == src.length())) {
    args.add(new Object[]{src, dst});
}
```

PaulSandoz (commented):
Thanks for your patience with my review comments. This is looking much better. Just a few minor comments.

PaulSandoz (approved):

sviswa7 (approved):

sviswa7:
@jatin-bhateja The patch looks good to me.

PR #41, opened 2021-06-14
8268663: Crash when guards contain boolean expression

lahodaj:
When Flow analyzes boolean expressions, (un-)initialized are split into (un-)initialized when true and when false. When analyzing guarded patterns, we need to handle these split values. As the only way the case's body/statements can be executed is if the case's pattern matches, we need to use the initialized values when true.

---------
### Progress

### Issue
 * [JDK-8268663](https://bugs.openjdk.java.net/browse/JDK-8268663): Crash when guards contain boolean expression

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/41/head:pull/41` \
`$ git checkout pull/41`

`$ git checkout pull/41` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/41/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 41`

`$ git pr show -t 41`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/41.diff">https://git.openjdk.java.net/jdk17/pull/41.diff</a>

</details>

### DIFF
--- a/test/langtools/tools/javac/patterns/Guards.java
+++ b/test/langtools/tools/javac/patterns/Guards.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -23,7 +23,7 @@
 
 /*
  * @test
- * @bug 8262891
+ * @bug 8262891 8268663
  * @summary Check guards implementation.
  * @compile --enable-preview -source ${jdk.version} Guards.java
  * @run main/othervm --enable-preview Guards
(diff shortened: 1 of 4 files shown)


test/langtools/tools/javac/patterns/Guards.java
```diff
@@ -84,6 +92,31 @@ String testBooleanSwitchExpression(Object o) {
         }
     }
 
+    String typeGuardIfTrueSwitchStatement(Object o) {
+        Object o2 = "";
+        switch (o) {
+            case Integer i && i == 0 && i < 1 && o2 instanceof String s: o = s + String.valueOf(i); return "true";
```
vicente-romero-oracle, line 98: could you please add tests with `||` in the condition?

lahodaj: Thanks, Vicente. I've added tests with `||` in https://github.com/openjdk/jdk17/pull/41/commits/d9e902b02b8a9916954fe4d2b665d4eaee3fb352 . One thing to note is that the tests are a bit different when writing something like: `case String s && s.isEmpty() || o2 instanceof Number n ->` `n` is not available here, as there is a path on which `o2` is not a `Number`.

vicente-romero-oracle (approved):
looks good

PR #45, opened 2021-06-14
8268566: java/foreign/TestResourceScope.java timed out

mcimadamore:
This patch contains another minor tweak to TestResourceScope as we have seen this test timing out at least once (on arm64).

I realized that some of the logic recently introduced in the test could lead to the test waiting forever: for instance, if all threads are executed right away before the main thread gets a chance to do anything, we'd be in a state where the lock count is zero, and no other thread will update it. I've removed the while loop - as that's not essential to make sure that the test works (what we want is to trigger a close operation that is concurrent with respect some acquire/release).

I've also lowered the number of threads - which was using 10K - that was good for stress testing, but it's not a good number in more realistic conditions.

---------
### Progress

### Issue
 * [JDK-8268566](https://bugs.openjdk.java.net/browse/JDK-8268566): java/foreign/TestResourceScope.java timed out

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/45/head:pull/45` \
`$ git checkout pull/45`

`$ git checkout pull/45` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/45/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 45`

`$ git pr show -t 45`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/45.diff">https://git.openjdk.java.net/jdk17/pull/45.diff</a>

</details>

### DIFF
--- a/test/jdk/java/foreign/TestResourceScope.java
+++ b/test/jdk/java/foreign/TestResourceScope.java
@@ -46,7 +46,7 @@ import java.util.stream.IntStream;
 
 public class TestResourceScope {
 
-    final static int N_THREADS = 10000;
+    final static int N_THREADS = 100;
 
     @Test(dataProvider = "cleaners")
     public void testConfined(Supplier<Cleaner> cleanerSupplier) {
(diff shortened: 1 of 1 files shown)


JornVernee (approved):
LGTM

PR #65, opened 2021-06-15
8268265:  MutableSpaceUsedHelper::take_sample() hits assert(left >= right) failed: avoid overflow

kimbarrett:
Please review this change to PerfData-based heap usage reporting by SerialGC
and ParallelGC.

For both of these collectors, the periodic heap usage sampling and reporting
by the Watcher thread involves sampling the difference between each space's
current allocation pointer and its base pointer.  But the Watcher thread
runs during GC safepoints, which means that sampling may occasionally occur
while the GC is in the process of updating the space being sampled.  This
can result in a sample with a bogus value or even an assertion failure after
JDK-8260046.

Other collectors seem to avoid this by sampling and recording the heap usage
value at specific synchronous times rather than asynchronously by the
Watcher thread.

There isn't a particularly good place for synchronous updates for ParallelGC
and SerialGC, so this change doesn't do so.  Instead, it makes the existing
asynchronous updates safe from interference by an in-progress GC.  This is
done by having the update done under the Heap_lock.  Introducing a more
fine-grained lock has the same problem as doing synchronous updates.  Rather
than possibly blocking the sampling Watcher thread for the duration of a GC,
it tries to lock the Heap_lock and skips sampling if it's not available.

Testing:
mach5 tier1-3, 5, 7 (3,5,7 are where focused Serial and ParallelGC testing is done).

---------
### Progress

### Issue
 * [JDK-8268265](https://bugs.openjdk.java.net/browse/JDK-8268265): MutableSpaceUsedHelper::take_sample() hits assert(left >= right) failed: avoid overflow

### Reviewers
 * [Thomas Schatzl](https://openjdk.java.net/census#tschatzl) (@tschatzl - **Reviewer**) ⚠️ Review applies to a04516e8d2ca3a37e10b725320efa6d89c943217
 * [Ivan Walulya](https://openjdk.java.net/census#iwalulya) (@walulyai - Committer) ⚠️ Review applies to a04516e8d2ca3a37e10b725320efa6d89c943217

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/65/head:pull/65` \
`$ git checkout pull/65`

`$ git checkout pull/65` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/65/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 65`

`$ git pr show -t 65`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/65.diff">https://git.openjdk.java.net/jdk17/pull/65.diff</a>

</details>


tschatzl (approved):
Lgtm.

walulyai (approved):
lgtm!

kimbarrett:
Thansk @tschatzl and @walulyai for reviews.

PR #78, opened 2021-06-16
8268871: Adjust javac to updated exhaustiveness specification

lahodaj:
The current updates spec for pattern matching in switch:
http://cr.openjdk.java.net/~gbierman/jep406/jep406-20210608/specs/patterns-switch-jls.html

tweaks the definition of exhaustive switches so that it include handling special types, like intersection types, sealed subclasses of sealed types, like:
```
sealed abstract class A permits B, C {}
final class B extends A {}
sealed abstract class C extends A permits D, E {}
final class D extends C {}
final class E extends C {}

public class Test {
  public static void main(String[] args) {
     A a = new D();
     System.out.println(
        switch (a) {
           case B x -> "B";
           case D x -> "D";
           case E x -> "E";
        });
  }
}
```

etc. This patch attempts to tweak javac to work according to this specification. It builds covered types based on what types are covered, i.e. if all subtypes of a sealed type are covered, the sealed type itself is covered, and this is done transitively, so that in the example above, `D` and `E` will mark `C` as covered, and then `B` and `C` will mark `A` as covered, and as the type if `a` is `A`, it means the switch is exhaustive.

---------
### Progress

### Issue
 * [JDK-8268871](https://bugs.openjdk.java.net/browse/JDK-8268871): Adjust javac to updated exhaustiveness specification

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/78/head:pull/78` \
`$ git checkout pull/78`

`$ git checkout pull/78` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/78/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 78`

`$ git pr show -t 78`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/78.diff">https://git.openjdk.java.net/jdk17/pull/78.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
@@ -30,7 +30,7 @@ package com.sun.tools.javac.comp;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.function.Predicate;
+import java.util.stream.StreamSupport;
 
 import com.sun.source.tree.LambdaExpressionTree.BodyKind;
 import com.sun.tools.javac.code.*;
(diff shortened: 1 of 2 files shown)


test/langtools/tools/javac/patterns/Exhaustiveness.java
```diff
@@ -640,4 +597,131 @@ private int test(S obj) {
...
+                   }
+               }
+               """);
+    }
+
+    @Test
+    public void testExhaustiveIntersection(Path base) throws Exception {
+        doTest(base,
+               new String[]{"""
+                            package lib;
```
vicente-romero-oracle, line 642: it will be nice to have additional tests with conditional patterns (total, with a condition that is true, and not) to stress all the possible code paths

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Flow.java
```diff
@@ -723,47 +723,91 @@ public void visitSwitchExpression(JCSwitchExpression tree) {
...
+                            if (hasAll) {
+                                if (covered.add(sym.owner)) {
+                                    modified = true;
+                                    continue OUTER;
+                                }
+                            }
+                        }
+
+                        case TYP -> {
+                            for (Type sup : types.directSupertypes(sym.type)) {
```
vicente-romero-oracle, line 773: just an observation, once a symbol has been processed, there is no need to process it again so you could keep a set of visited symbols to avoid visiting them again. Not a big deal performance wise unless we are switching on a sealed hierarchy with a lot of members I guess > 50 but we know what automatic code generators are capable of :)

test/langtools/tools/javac/patterns/Exhaustiveness.java
lahodaj: Thanks Vicente! I've updated the patch based on your comments in https://github.com/openjdk/jdk17/pull/78/commits/a7a61692dcf3fbc8d24a5cf7c881ec635d654f62 .

What do you think?

Thanks!

vicente-romero-oracle (approved):
lgtm

PR #79, opened 2021-06-16
8268714: [macos-aarch64] 7 java/net/httpclient/websocket tests failed

dfuch:
Hi, 

Please find below a test-only change to fix some intermittent failures observed with the httpclient/websocket tests:
these tests intermittently and randomly fail with ENOMEM ("No buffer space available").

Some machines in our CI seem to allow a higher level of concurrency while being (maybe) configured with lower system resources (such as available buffer space for the TCP stack).

Some of the httpclient/websocket tests attempt to fill the sockets buffers in order to assert some conditions when the buffers are full and writing is paused. When the test process terminates, this leaves behind TCP sockets in the TIME_WAIT state that still hold system buffer resources in case retransmission is needed. When several such tests are run this ends up causing random "No buffer space available" errors on other tests (including these tests themselves) running concurrently or shortly after on the same machine.

This change implements a few tricks to alleviate the situation:
 - configure the tests with smaller send buffers on the client side and receive buffers on the server side, in order to limit how much buffer space is consumed by the test.
 - when the not-reading server is closed, and before the accepted socket is closed, read all available data off the socket buffer in order to free up the buffer space that the test has consumed before closing the socket.
 - in some tests that create a large number of HttpClients, limit the number of clients created in shared client mode, and add a call to System.gc() and a small pause to give time for gc to collect the old clients which are no longer referenced. 
 
 With these changes, I have run the HttpClient tests 200 times on the problematic machines without observing any failures (where previously there was at least a couple of failures per 50 runs). I also ran tier1 once, and tier2 twice and the results came clean.
 
 I am therefore claiming success (even if it might prove temporary ;-) )
 
If these failures come back to haunt the CI again after this fix, a further remediation policy could be to put the httpclient/websocket directory in exclusive test execution mode (in TEST.root) - this seems to work too - but cleaning up garbage in the tests themselves seems preferable.

---------
### Progress

### Issue
 * [JDK-8268714](https://bugs.openjdk.java.net/browse/JDK-8268714): [macos-aarch64] 7 java/net/httpclient/websocket tests failed

### Reviewers
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - **Reviewer**)
 * [Michael McMahon](https://openjdk.java.net/census#michaelm) (@Michael-Mc-Mahon - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/79/head:pull/79` \
`$ git checkout pull/79`

`$ git checkout pull/79` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/79/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 79`

`$ git pr show -t 79`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/79.diff">https://git.openjdk.java.net/jdk17/pull/79.diff</a>

</details>

### DIFF
--- a/test/jdk/java/net/httpclient/ISO_8859_1_Test.java
+++ b/test/jdk/java/net/httpclient/ISO_8859_1_Test.java
@@ -278,6 +278,7 @@ public class ISO_8859_1_Test implements HttpServerAdapters {
 
     @Test(dataProvider = "variants")
     public void test(String uri, boolean sameClient) throws Exception {
+        checkSkip();
         System.out.println("Request to " + uri);
 
         HttpClient client = newHttpClient(sameClient);
--- a/test/jdk/java/net/httpclient/websocket/DummyWebSocketServer.java
+++ b/test/jdk/java/net/httpclient/websocket/DummyWebSocketServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2019, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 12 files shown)


ChrisHegarty (approved):
Looks good to me. ( a pragmatic approach to the issue at hand - resolve intermittently failing tests )

Michael-Mc-Mahon (approved):
LGTM

PR #81, opened 2021-06-16
8268766: Desugaring of pattern matching enum switch should be improved

lahodaj:
Currently, an enum switch with patterns is desugared in a very non-standard, and potentially slow, way. It would be better to use the standard `typeSwitch` bootstrap to classify the enum constants. The bootstrap needs to accept enum constants as labels in order to allow this. A complication is that if an enum constant is missing, that is not an incompatible change for the switch, and the switch should simply work as if the case for the missing constant didn't exist. So, the proposed solution is to have a new bootstrap `enumSwitch` that accepts `String`s in place of the enum constants, and will internally convert them to the appropriate enum constants, and then it will find the proper case similarly to `typeSwitch`.

How does this look?

---------
### Progress

### Issue
 * [JDK-8268766](https://bugs.openjdk.java.net/browse/JDK-8268766): Desugaring of pattern matching enum switch should be improved

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - **Reviewer**) ⚠️ Review applies to ca79b2bdb116789944f10bcba3455872f18ed021

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/81/head:pull/81` \
`$ git checkout pull/81`

`$ git checkout pull/81` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/81/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 81`

`$ git pr show -t 81`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/81.diff">https://git.openjdk.java.net/jdk17/pull/81.diff</a>

</details>


lahodaj:
I was thinking of having a new bootstrap method for enums, but then (in a discussion with Maurizio), it turned out it might be better to reuse `typeSwitch` for consistency and orthogonality. The usecase is not really much different from e.g. pattern switches over Strings (or Integers). One can write something like:
```
         String sel = "";
         switch (sel) {
             case "a" -> {}
             case String s && s.length() < 5 -> {}
             case "toolong" -> {}
             case String s -> {}
         }
```

And javac will use `typeSwitch` for that (and I think it is more or less necessary to mix constants and patterns in the static arguments of the bootstrap method, as the switch may mix constants and patterns with guards in any order). We can do something similar with enums:
```
         E sel = null;
         switch (sel) {
             case A -> {}
             case E e && "B".equals(e.name()) -> {}
             case C -> {}
             case E e -> {}
         }
```

Does not seem much different, the only change is that the constants are enum constants, not String constants (which brings some challenges - like the possibility that the enum constant cannot be resolved).

But I can create a separate bootstrap method, if strongly preferred.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +169,114 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
         return labels.length;
     }
 
+    /**
+     * Bootstrap method for linking an {@code invokedynamic} call site that
```
mcimadamore, line 173: This should be made clearer - e.g. the first argument is of type `Class` and represents the enum we want to switch on. The remaining constants should be of type `String`, the names of the various constants.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +169,114 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+     * implements a {@code switch} on a target of an enum type.  The static
+     * arguments are an array of case labels which must be non-null and of type
+     * {@code String} (representing enum constants) or {@code Class}.
+     * <p>
+     * The type of the returned {@code CallSite}'s method handle will have
```
mcimadamore, line 178:
```suggestion
     * The returned {@code CallSite}'s method handle will have
```

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +169,114 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+     * a return type of {@code int}.   It has two parameters: the first argument
```
mcimadamore, line 179:
```suggestion
     * a return type of {@code int}.   It has two parameters: the first argument
```
```suggestion
     * a return type of {@code int} and accepts two parameters: the first argument
```

test/langtools/tools/javac/patterns/EnumTypeChanges.java
```diff
@@ -0,0 +1,80 @@
...
+        if (!Objects.equals(o1, o2)) {
+            throw new AssertionError();
+        }
+    }
+}
+
+enum EnumTypeChangesEnum {
+    A,
+    B;
+}
```
mcimadamore, line 80: missing newline here

test/langtools/tools/javac/patterns/EnumTypeChanges2.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+enum EnumTypeChangesEnum {
+    A,
+    C;
+}
```
mcimadamore, line 27: missing newline here

mcimadamore (commented):
Very good piece of work. I like all the code that can be removed because of this.

I assume that the new code only kicks in if there's at least a pattern in the switch, otherwise we fallback to legacy translation (meaning that compiling with source < 17 is still ok), right?

I left some comments to help and clarify the javadoc text of the enum BSM.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +169,114 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+        requireNonNull(labels);
+
+        labels = labels.clone();
+
+        Class<?> enumClass = invocationType.parameterType(0);
+        labels = Stream.of(labels).map(l -> convertEnumConstants(lookup, enumClass, l)).toArray();
+
+        MethodHandle target =
+                MethodHandles.insertArguments(DO_ENUM_SWITCH, 2, (Object) labels);
+        target = MethodHandles.explicitCastArguments(target, invocationType);
```
forax, line 238: why explicitCast is used here instead of the classical cast asType() ?

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +169,114 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+     * or if {@code labels} contains an element that is not of type {@code String} or
+     * {@code Class}.
+     * @throws Throwable if there is any error linking the call site
+     * @jvms 4.4.6 The CONSTANT_NameAndType_info Structure
+     * @jvms 4.4.10 The CONSTANT_Dynamic_info and CONSTANT_InvokeDynamic_info Structures
+     */
+    public static CallSite enumSwitch(MethodHandles.Lookup lookup,
+                                      String invocationName,
+                                      MethodType invocationType,
+                                      Object... labels) throws Throwable {
```
forax, line 222: Is it not better to take a Class and a String... as separated parameters instead of taking Object... and doing the conversion to a Class and an array of String later in Java

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
lahodaj: That is not quite what the labels represent. The target Enum type is inferred from the bootstrap method's invocation MethodType. (Alternatively, we can add a new Class parameter to the bootstrap method.)

For the labels, please consider this switch:
```
         E sel = null;
         switch (sel) {
             case A -> {}
             case E e && "B".equals(e.name()) -> {}
             case C -> {}
             case E e -> {}
         }
```

The patterns and the constants are mixed, and the order needs to be represented somehow in the labels array, so that the switch will classify the input correctly. The method in this proposal will use `{"A", E.class, "C", E.class}` as the labels array (which is mostly consistent with the `typeSwitch` method), but we could use different encodings if needed.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
lahodaj: This is to represent cases like:
```
         E sel = null;
         switch (sel) {
             case A -> {}
             case E e && "B".equals(e.name()) -> {}
             case C -> {}
             case E e -> {}
         }
```

The method needs to know which cases represent constants and which represent patterns (even though the primary type of all the patterns will be the enum type), so we cannot easily put the `Class` first (or elide it), and then a `String...`, unless we represent the patterns in the `String...` array somehow.

lahodaj:
> Very good piece of work. I like all the code that can be removed because of this.

Thanks!

> 
> I assume that the new code only kicks in if there's at least a pattern in the switch, otherwise we fallback to legacy translation (meaning that compiling with source < 17 is still ok), right?

Yes, it is only for pattern matching switches. Traditional switches are still desugared in the way they were.

> 
> I left some comments to help and clarify the javadoc text of the enum BSM.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
forax: Ok got it,
At some point, we will have to represent patterns either as String and parse them or as Condy of condy if we want a more structured recursive way.

mcimadamore: Of course - I missed it! Thanks for the explanation.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -175,8 +175,8 @@ private static int doTypeSwitch(Object target, int startIndex, Object[] labels)
      * arguments are an array of case labels which must be non-null and of type
```
mcimadamore, line 175: This sentence can still be improved and made cleared. Example:

> The static arguments are used to encode the case labels associated to the `switch` construct, where each label can be encoded as a `String` (e.g. to represent an enum constant), or, alternatively, as a `Class` (e.g. to represent a type test pattern whose type is an enum type).

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -175,8 +175,8 @@ private static int doTypeSwitch(Object target, int startIndex, Object[] labels)
...
      * {@code String} (representing enum constants) or {@code Class}.
      * <p>
-     * The type of the returned {@code CallSite}'s method handle will have
-     * a return type of {@code int}.   It has two parameters: the first argument
+     * The returned {@code CallSite}'s method handle will have
+     * a return type of {@code int} and accepts two parameters: the first argument
```
mcimadamore, line 179: By writing the javadoc text above, it came to mind that, perhaps, some validation is in order for the static arguments. For instance:

* the enum type of a Class parameter doesn't match that of the BSM
* the static arguments contain more than one Class parameter (I think even if they are all of the same "correct" type, that's bogus, right?)

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
lahodaj: Thanks. I've updated the javadoc and improved the validation a little, checking the `Class` objects provided.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -171,9 +171,11 @@ private static int doTypeSwitch(Object target, int startIndex, Object[] labels)
 
     /**
      * Bootstrap method for linking an {@code invokedynamic} call site that
-     * implements a {@code switch} on a target of an enum type.  The static
-     * arguments are an array of case labels which must be non-null and of type
-     * {@code String} (representing enum constants) or {@code Class}.
+     * implements a {@code switch} on a target of an enum type. The static
+     * arguments are used to encode the case labels associated to the switch
```
mcimadamore, line 175: `String` and `Class` should appear in code blocks perhaps, or link tags? Also, I think this text could be improved by splitting the sentence by using a bullet list:

```
The static arguments are used to encode the case labels associated to the switch
construct, where each label can be encoded in two ways:
* as a String value, which represents the name of the enum constant associated with the label
* as a Class value, which represents the enum type associated with a type test pattern
```

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -211,7 +213,7 @@ private static int doTypeSwitch(Object target, int startIndex, Object[] labels)
      * invocation type is not a method type of first parameter of an enum type,
      * second parameter of type {@code int} and with {@code int} as its return type,
      * or if {@code labels} contains an element that is not of type {@code String} or
-     * {@code Class}.
+     * {@code Class} of the target enum type.
      * @throws Throwable if there is any error linking the call site
      * @jvms 4.4.6 The CONSTANT_NameAndType_info Structure
```
mcimadamore, line 218: Not sure what we do for other bootstraps, but I wonder if the throws clause here should be more specific - for instance the LambdaMetafactory tells when an exception is gonna be triggered. Here it seems like there are a couple of cases which can be called out, where we know that linkage is going to fail.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
lahodaj: Thanks! I've (tried to) updated the javadoc as suggested.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +169,123 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+     * second parameter of type {@code int} and with {@code int} as its return type,
+     * or if {@code labels} contains an element that is not of type {@code String} or
+     * {@code Class} of the target enum type.
+     * @jvms 4.4.6 The CONSTANT_NameAndType_info Structure
+     * @jvms 4.4.10 The CONSTANT_Dynamic_info and CONSTANT_InvokeDynamic_info Structures
+     */
+    public static CallSite enumSwitch(MethodHandles.Lookup lookup,
+                                      String invocationName,
+                                      MethodType invocationType,
+                                      Object... labels) throws NullPointerException,
```
PaulSandoz, line 227: I don't think there are any benefits to declaring the runtime exceptions in the method declaration.

Other bootstrap methods declare a checked exception when certain linkage constraints are violated (and the line between whats an `IllegalArgumentException` or an explicit linkage checked exception can be blurry). In your case, given the simplicity i think what you have is ok. We could refine later with a specific checked exception for switch linkage violations.

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +169,123 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+    }
+
+    private static <E extends Enum<E>> Object convertEnumConstants(MethodHandles.Lookup lookup, Class<?> enumClassTemplate, Object label) {
+        if (label == null) {
+            throw new IllegalArgumentException("null label found");
+        }
+        Class<?> labelClass = label.getClass();
+        if (labelClass == Class.class) {
+            if (label != enumClassTemplate) {
+                throw new IllegalArgumentException("illegal Class label: " + label);
```
PaulSandoz, line 256: Can we refine the message to state the class label is not equal to the enum class that is the target of the switch? LIkewise in the `else` block to state the label is not of class String or the target enum class.

PaulSandoz (approved):
Bootstrap method looks. At first i thought why not let the class label just be an actual `Class.class` instance, signaling that target enum class should be used, but then thought perhaps it's better to be more literal: clarity of inputs, unlikely label lists will be shared, and further it might be easier to adapt them.

PaulSandoz (approved):

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +168,124 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+     *       from the target's class; or</li>
+     *   <li>the element is of type {@code String} and equals to the target
+     *       enum constant's {@link Enum#name()}.</li>
+     * </ul>
+     * <p>
+     * If no element in the {@code labels} array matches the target, then
+     * the method of the call site return the length of the {@code labels} array.
+     *
+     * @param lookup Represents a lookup context with the accessibility
+     *               privileges of the caller.  When used with {@code invokedynamic},
```
mcimadamore, line 205:
```suggestion
     *               privileges of the caller. When used with {@code invokedynamic},
```

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +168,124 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+     * @param invocationName unused
+     * @param invocationType The invocation type of the {@code CallSite} with two parameters,
+     *                       an enum type, an {@code int}, and {@code int} as a return type.
+     * @param labels case labels - {@code String} constants and {@code Class} instances,
+     *               in any combination
+     * @return a {@code CallSite} returning the first matching element as described above
+     *
+     * @throws NullPointerException if any argument is {@code null}
+     * @throws IllegalArgumentException if any element in the labels array is null, if the
+     * invocation type is not a method type of first parameter of an enum type,
```
mcimadamore, line 216:
```suggestion
     * invocation type is not a method type whose first parameter type is an enum type,
```

src/java.base/share/classes/java/lang/runtime/SwitchBootstraps.java
```diff
@@ -167,4 +168,124 @@ private static int doSwitch(Object target, int startIndex, Object[] labels) {
...
+     * second parameter of type {@code int} and with {@code int} as its return type,
```
mcimadamore, line 217:
```suggestion
     * second parameter of type {@code int} and whose return type is {@code int},
```

mcimadamore (approved):
Looks good - added minor cosmetic comments

mcimadamore (approved):

lahodaj:
There turned out to be a bug in the patch: considering a switch over enum like:
```
    E e = ...;
    switch (e) {
        case Object o -> {}
    }
```

(Or, even worse, `case Object o && <guard> ->`), the patch will generate `Object.class` as a static parameter to the `enumSwitch` bootstrap method. But the method (rightfully) requires the enum class as the static parameter. https://github.com/openjdk/jdk17/pull/81/commits/d970402e969d76a017cdfdcbc6556f6d9a9f3bfa tweaks the code generation to produce `E.class` instead of `Object.class`.

mcimadamore (approved):
Good catch - should we also add test for when an enum implements an interface?

```
enum Foo implements A { ... }

Foo foo = ...
switch(foo) {
   case A a: ...
}
```

I think the patch you have should work since you are using subtyping.

Another thing to consider: should we be worried about type erasure turning the selector type into some weird, unexpected type? I don't think that should be the case, given that if a type variable has a bound E, where E is an enum, the erasure should always be E (as E is a "class" type, so should win, even in intersections).

mcimadamore (approved):
Looks good

PR #88, opened 2021-06-17
8268672: C2: assert(!loop->is_member(u_loop)) failed: can be in outer loop or out of both loops only

rwestrel:
This is caused by the fix for 8263303 in which I changed the code of
loop cloning of a strip mined loop so nodes that are pinned in the
outer loop but not referenced by the loop's safepoint are moved out of
the loop before cloning:

https://github.com/openjdk/jdk/commit/d4377afb999f4f03d384ded97771c83ea1c1f513#diff-7b82624b78127158abbce6835eeba196bd062aee59512ec2d4e4c8c7d681573bR1980

The assert failure here is caused by a node that has its control input
set to the inner loop but is assigned control by PhaseIdealLoop out of
the inner loop. The assert fires because the change from 8263303
affects the condition that the assert checks. I still think, in
principle, the fix from 8263303 is the good one so I propose to simply
reorganize the change from 8263303 so the assert doesn't wrongly fire.

---------
### Progress

### Issue
 * [JDK-8268672](https://bugs.openjdk.java.net/browse/JDK-8268672): C2: assert(!loop->is_member(u_loop)) failed: can be in outer loop or out of both loops only

### Reviewers
 * [Nils Eliasson](https://openjdk.java.net/census#neliasso) (@neliasso - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/88/head:pull/88` \
`$ git checkout pull/88`

`$ git checkout pull/88` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/88/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 88`

`$ git pr show -t 88`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/88.diff">https://git.openjdk.java.net/jdk17/pull/88.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/loopopts.cpp
+++ b/src/hotspot/share/opto/loopopts.cpp
@@ -1980,7 +1980,10 @@ static void clone_outer_loop_helper(Node* n, const IdealLoopTree *loop, const Id
       Node* c = phase->get_ctrl(u);
       IdealLoopTree* u_loop = phase->get_loop(c);
       assert(!loop->is_member(u_loop), "can be in outer loop or out of both loops only");
-      if (outer_loop->is_member(u_loop) || (u->in(0) != NULL && outer_loop->is_member(phase->get_loop(u->in(0))))) {
+      if (outer_loop->is_member(u_loop) ||
+          // nodes pinned with control in the outer loop but not referenced from the safepoint must be moved out of
+          // the outer loop too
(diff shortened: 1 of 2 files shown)


vnkozlov (approved):
Looks good.

neliasso (approved):
Looks good!

rwestrel:
@vnkozlov @neliasso thanks for the reviews

PR #91, opened 2021-06-17
8254571: Erroneous generic type inference in a lambda expression with a checked exception

vicente-romero-oracle:
Hi,

The compiler is issuing an error for code like:

```
class JDK8254571 {
    public static void test() {
        outer(nested(x -> mightThrow()));
    }
    
    static <A> void outer(Object o) {}

    static <B, C, E extends Throwable> B nested(Fun<C,E> fun) {
        return null;
    }

    interface Fun<X, Y extends Throwable> {
        void m(X t) throws Y;
    }

    static void mightThrow() throws Exception {}
} 
```

there are several things at play here:

1) for performance reasons we decided to minimize the number of inference variables popped up when inferring nested generic method invocations, see Infer::roots
2) when an expression is stuck (like in this case this implicit lambda) we decide which inference variables should be present in the related inference context and thrown variables were not part of the equation

So there are three hypothesis here:
1) there is a bug in the code that determines which variables are passed up to the nested inference context, meaning that it is passing less variables than the bare minimum needed
2) there is a bug in determining which inference variables should be extracted from stuck expressions
3) both are buggy

The root cause of this bug is that some types escape from Attr with non instantiated type variables in them. I modified the test case to double check if the issue was not related to lambdas, but then the test passed so in the example above I added:

```
     static <X, Y extends Throwable> Fun<X, Y> mightThrow2() throws Y { return null; }
```

which simulates the lambda and the compiler has no issues compiling this invocation:

```
    outer(nested(mightThrow2()));
```

so summing up, after double checking the logic of the code that minimized the inference context I realized that the issue is not there. The code is sound and its output only depend on its input, so the issue should be on determining the input. So IMO the issue is in what inference variables we extract from stuck expressions, which is one of the inputs to Infer::roots. Looking for what variables we extract from stuck expressions, I realized that the code currently ignores type variables in the throws clause and this is why my proposed fix is addressing this issue.

Comments?

---------
### Progress

### Issue
 * [JDK-8254571](https://bugs.openjdk.java.net/browse/JDK-8254571): Erroneous generic type inference in a lambda expression with a checked exception

### Reviewers
 * [Jan Lahoda](https://openjdk.java.net/census#jlahoda) (@lahodaj - **Reviewer**) ⚠️ Review applies to 343943b9888a7dd7bea106c46906bf228036c518
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to 343943b9888a7dd7bea106c46906bf228036c518

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/91/head:pull/91` \
`$ git checkout pull/91`

`$ git checkout pull/91` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/91/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 91`

`$ git pr show -t 91`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/91.diff">https://git.openjdk.java.net/jdk17/pull/91.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/DeferredAttr.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/DeferredAttr.java
@@ -1217,6 +1217,7 @@ public class DeferredAttr extends JCTree.Visitor {
                     freeArgVars.nonEmpty()) {
                 stuckVars.addAll(freeArgVars);
                 depVars.addAll(inferenceContext.freeVarsIn(descType.getReturnType()));
+                depVars.addAll(inferenceContext.freeVarsIn(descType.getThrownTypes()));
             }
             scanLambdaBody(tree, descType.getReturnType());
         }
@@ -1238,6 +1239,7 @@ public class DeferredAttr extends JCTree.Visitor {
                     tree.getOverloadKind() != JCMemberReference.OverloadKind.UNOVERLOADED) {
                 stuckVars.addAll(freeArgVars);
                 depVars.addAll(inferenceContext.freeVarsIn(descType.getReturnType()));
+                depVars.addAll(inferenceContext.freeVarsIn(descType.getThrownTypes()));
             }
         }
 
--- a/test/langtools/tools/javac/lambda/considerExceptionTVarInStuckExprs/ConsiderExceptionTVarsInStuckExprs.java
+++ b/test/langtools/tools/javac/lambda/considerExceptionTVarInStuckExprs/ConsiderExceptionTVarsInStuckExprs.java
@@ -32,6 +32,7 @@ class ConsiderExceptionTVarsInStuckExprs {
 
     public static void test() {
         outer(nested(x -> mightThrow()));
+        outer(nested(ConsiderExceptionTVarsInStuckExprs::mightThrow2));
     }
 
     static <A> void outer(Object o) {}
@@ -45,4 +46,5 @@ class ConsiderExceptionTVarsInStuckExprs {
     }
 
     static void mightThrow() throws Exception {}
+    static <C> void mightThrow2(C c) throws Exception {}
 }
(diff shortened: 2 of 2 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/DeferredAttr.java
```diff
@@ -1238,6 +1239,7 @@ public void visitReference(JCMemberReference tree) {
                     tree.getOverloadKind() != JCMemberReference.OverloadKind.UNOVERLOADED) {
                 stuckVars.addAll(freeArgVars);
                 depVars.addAll(inferenceContext.freeVarsIn(descType.getReturnType()));
+                depVars.addAll(inferenceContext.freeVarsIn(descType.getThrownTypes()));
```
vicente-romero-oracle, line 1242: did the same for method references for completion but I couldn't find a test case that reproduced the issue with a method reference

vicente-romero-oracle:
ping

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/DeferredAttr.java
lahodaj: How about:
```
outer(nested(ConsiderExceptionTVarsInStuckExprs::mightThrow2));
...
static <C> void mightThrow2(C c) throws Exception {}
```

lahodaj (approved):
Looks reasonable to me. Not sure if I can foresee the consequences, though.

mcimadamore (approved):
I believe the fix is good. We probably overlooked free type variables in the thrown clause (as they are relatively rare) when doing stuck expression computation. These variables are part of the bound set used by JLS chapter 18, so dropping them on the floor (as current code does) is not a good option. This patch rectifies that.

To test with method references, you would need the method reference to be "non-exact": only in that case will the method reference be considered stuck in the same way the implicit lambda is, I believe.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/DeferredAttr.java
vicente-romero-oracle: yep that one makes the trick, thanks

vicente-romero-oracle:
thanks guys for the review

PR #92, opened 2021-06-17
JDK-8268972: Add default impl for recent new Reporter.print method

jonathan-gibbons:
Please review a simple fix to add a default implementation for a recent new method on the Reporter interface. (It was an oversight that the default implementation was not provided in the original work.)

---------
### Progress

### Issue
 * [JDK-8268972](https://bugs.openjdk.java.net/browse/JDK-8268972): Add default impl for recent new Reporter.print method

### Reviewers
 * [Pavel Rappo](https://openjdk.java.net/census#prappo) (@pavelrappo - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/92/head:pull/92` \
`$ git checkout pull/92`

`$ git checkout pull/92` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/92/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 92`

`$ git pr show -t 92`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/92.diff">https://git.openjdk.java.net/jdk17/pull/92.diff</a>

</details>

### DIFF
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Reporter.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Reporter.java
@@ -94,8 +94,8 @@ public interface Reporter {
      * The positions are all 0-based character offsets from the beginning of content of the file.
      * The positions should satisfy the relation {@code start <= pos <= end}.
      *
-     * @implNote
-     * This implementation throws {@code UnsupportedOperationException}.
+     * @implSpec
+     * This implementation always throws {@code UnsupportedOperationException}.
      * The implementation provided by the {@code javadoc} tool to
      * {@link Doclet#init(Locale, Reporter) initialize} a doclet
      * overrides this implementation.
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Reporter.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Reporter.java
@@ -94,6 +94,12 @@ public interface Reporter {
      * The positions are all 0-based character offsets from the beginning of content of the file.
      * The positions should satisfy the relation {@code start <= pos <= end}.
      *
+     * @implNote
+     * This implementation throws {@code UnsupportedOperationException}.
+     * The implementation provided by the {@code javadoc} tool to
+     * {@link Doclet#init(Locale, Reporter) initialize} a doclet
+     * overrides this implementation.
+     *
      * @param kind    the kind of diagnostic
      * @param file    the file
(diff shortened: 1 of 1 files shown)


src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Reporter.java
```diff
@@ -94,6 +94,12 @@
      * The positions are all 0-based character offsets from the beginning of content of the file.
      * The positions should satisfy the relation {@code start <= pos <= end}.
      *
+     * @implNote
```
pavelrappo, line 97: Should be `@implSpec` instead: both here and in the respective CSR.

src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Reporter.java
```diff
@@ -94,6 +94,12 @@
...
+     * This implementation throws {@code UnsupportedOperationException}.
```
pavelrappo, line 98: I'd suggest rephrasing this as follows: "This implementation always throws..." both here and in the respective CSR.

src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Reporter.java
```diff
@@ -94,6 +94,12 @@
...
+     * The implementation provided by the {@code javadoc} tool to
+     * {@link Doclet#init(Locale, Reporter) initialize} a doclet
+     * overrides this implementation.
```
pavelrappo, line 101: Overrides how?

pavelrappo (changes requested):
Synchronize with the respective CSR.

src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Reporter.java
jonathan-gibbons: ...per the documented spec.   I tried writing `The implementation .... always generates an appropriate diagnostic` but that's not true if an exception is thrown, and writing `The implementation ... does not throw UOE` seems weird.

pavelrappo: The difficulty you have with ironing out this wording wrinkle suggests that the behavior of the `Reporter` provided to `Doclet.init` should be described in `Doclet.init` rather than here. The same goes for other default methods of Reporter.

pavelrappo (approved):

PR #97, opened 2021-06-18
8268612: a few runtime/memory tests don't check exit code

iignatev:
Hi all,

could you please review this small patch that adds checks of exit code to a few `runtime/memory` tests?
from JBS:
> `ReserveMemory.java`, `ReadFromNoaccessArea.java` and `TestLargePagesFlags.java` tests spawn new JVMs but don't check their exit code which might lead to both type-I and type-II errors

in `ReadFromNoaccessArea.java`, the patch removes throwing of an exception in `DummyClassWithMainTryingToReadFromNoaccessArea` (child process), so we will get 0 as an exit code if/when we don't crash.

testing: `runtime/memory` tests on `{linux,windows,macosx}-x64`

Thanks,
-- Igor

---------
### Progress

### Issue
 * [JDK-8268612](https://bugs.openjdk.java.net/browse/JDK-8268612): a few runtime/memory tests don't check exit code

### Reviewers
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to ab9b807054117d27c934ad756e58db87f5c9242b
 * [Mikhailo Seledtsov](https://openjdk.java.net/census#mseledtsov) (@mseledts - Committer) ⚠️ Review applies to ab9b807054117d27c934ad756e58db87f5c9242b

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/97/head:pull/97` \
`$ git checkout pull/97`

`$ git checkout pull/97` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/97/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 97`

`$ git pr show -t 97`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/97.diff">https://git.openjdk.java.net/jdk17/pull/97.diff</a>

</details>

### DIFF
--- a/test/hotspot/jtreg/runtime/memory/ReadFromNoaccessArea.java
+++ b/test/hotspot/jtreg/runtime/memory/ReadFromNoaccessArea.java
@@ -62,6 +62,7 @@ public class ReadFromNoaccessArea {
     if (output.getStdout() != null && output.getStdout().contains("WB_ReadFromNoaccessArea method is useless")) {
       throw new SkippedException("There is no protected page in ReservedHeapSpace in these circumstance");
     }
+    output.shouldNotHaveExitValue(0);
     if (Platform.isWindows()) {
       output.shouldContain("EXCEPTION_ACCESS_VIOLATION");
     } else if (Platform.isOSX()) {
@@ -76,7 +77,6 @@ public class ReadFromNoaccessArea {
     // This method calls whitebox method reading from noaccess area
     public static void main(String args[]) throws Exception {
       WhiteBox.getWhiteBox().readFromNoaccessArea();
-      throw new Exception("Call of readFromNoaccessArea succeeded! This is wrong. Crash expected. Test failed.");
     }
   }
 
--- a/test/hotspot/jtreg/runtime/memory/ReserveMemory.java
+++ b/test/hotspot/jtreg/runtime/memory/ReserveMemory.java
@@ -59,6 +59,7 @@ public class ReserveMemory {
           "test");
 
     OutputAnalyzer output = new OutputAnalyzer(pb.start());
+    output.shouldNotHaveExitValue(0);
     if (Platform.isWindows()) {
       output.shouldContain("EXCEPTION_ACCESS_VIOLATION");
     } else if (Platform.isOSX()) {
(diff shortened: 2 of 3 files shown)


test/hotspot/jtreg/runtime/memory/ReadFromNoaccessArea.java
```diff
@@ -62,6 +62,7 @@ public static void main(String args[]) throws Exception {
     if (output.getStdout() != null && output.getStdout().contains("WB_ReadFromNoaccessArea method is useless")) {
       throw new SkippedException("There is no protected page in ReservedHeapSpace in these circumstance");
     }
+    output.shouldNotHaveExitValue(0);
```
dholmes-ora, line 65: This is redundant. You will never have a zero exit value if you match the crash messages or else throw the exception. Seeing the crash message implies the return code is not zero so checking is a waste of time IMO not a more rigorous check. If the test did not crash then we won't match the expected output and we will still fail.

test/hotspot/jtreg/runtime/memory/ReserveMemory.java
```diff
@@ -59,6 +59,7 @@ public static void main(String args[]) throws Exception {
           "test");
 
     OutputAnalyzer output = new OutputAnalyzer(pb.start());
+    output.shouldNotHaveExitValue(0);
```
dholmes-ora, line 62: Ditto - redundant.

dholmes-ora (changes requested):
Sorry Igor I don't think 2 of these changes are needed/warranted.

David

test/hotspot/jtreg/runtime/memory/ReadFromNoaccessArea.java
dholmes-ora: Sorry, please ignore my previous comment. If we inadvertently caused the call to exit with a zero return in case of a crash, this change would detect it.

dholmes-ora (approved):
Hi Igor,

Apologies for my previous review it was a bit brash.

A small change suggested in ReserveMemoryTest - you should also remove the throwing of the exception on a successful read as you did for  ReadFromNoAccess.

Thanks,
David

mseledts (approved):
Looks good to me.

iignatev:
Misha and David, thank you for your reviews!

/integrate

PR #106, opened 2021-06-20
8260684: vmTestbase/gc/gctests/PhantomReference/phantom002/TestDescription.java timed out

kimbarrett:
Please review this fix of
vmTestbase/gc/gctests/PhantomReference/phantom001/phantom001.java
to eliminate a race that can lead to the test very occasionally hanging and
eventually timing out.

When the referent being tested by a particular iteration is a finalizable
class, the test invoked eatMemory() repeatedly until the referent's
finalize() function set a flag, then called eatMemory() once more to cause
the reference to be cleared and notified.  The problem with this is that
eatMemory() might trigger its GCs before the finalization thread has dropped
the referent, so the referent is still live and the reference has not been
cleared and notified afterward.

The testing thread then proceeds to wait for the reference to be notified.
This usually doesn't cause a problem because other threads will trigger
further GCs, and the finalization thread will eventually drop the object.
However, if the test is nearing completion and this situation arises in the
last thread running, it will wait forever for the notification that isn't
coming.

While investigating this issue I did a bunch of code cleanup (for example,
there was some really badly formatted code), refactoring, and commenting.
As a result, a large fraction of the test has changed, though mostly in ways
that don't affect it's function.  The first of the two commits in the PR is
a whitespace-only cleanup.  Excluding it may make reviewing easier.

Testing:
Ran the phantom002 test a couple thousand times.

Using a modified version of the test, verified that the described problem
case actually occurs; that is, one eatMemory call is sometimes insufficient
to cause clearing of the reference because the referent is still live on the
finalization thread.

/label hotspot-gc

---------
### Progress

### Issue
 * [JDK-8260684](https://bugs.openjdk.java.net/browse/JDK-8260684): vmTestbase/gc/gctests/PhantomReference/phantom002/TestDescription.java timed out

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/106/head:pull/106` \
`$ git checkout pull/106`

`$ git checkout pull/106` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/106/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 106`

`$ git pr show -t 106`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/106.diff">https://git.openjdk.java.net/jdk17/pull/106.diff</a>

</details>

### DIFF
--- a/test/hotspot/jtreg/vmTestbase/gc/gctests/PhantomReference/phantom001/phantom001.java
+++ b/test/hotspot/jtreg/vmTestbase/gc/gctests/PhantomReference/phantom001/phantom001.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2004, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -60,6 +60,7 @@
 package gc.gctests.PhantomReference.phantom001;
 
 import java.lang.ref.*;
+import java.time.LocalTime;
 import nsk.share.gc.*;
 import nsk.share.gc.gp.*;
 import nsk.share.gc.gp.string.InternedStringProducer;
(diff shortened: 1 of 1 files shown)


kimbarrett:
Oops, I messed up the last change, and testing didn't find it.  Withdrawing this change.

PR #112, opened 2021-06-21
8269096: Add java.util.Objects.newIdentity method

RogerRiggs:
Add java.util.Objects.newIdentity to supply a unique object with identity.
This is a replacement code can be used today for the traditional new Object() idiom, which will be deprecated under Project Valhalla.
Refer to [JEP 401: Primitive Objects (Preview)](https://openjdk.java.net/jeps/401) for background.

---------
### Progress

### Integration blocker
&nbsp;⚠️ The change requires a CSR request to be approved.

### Issue
 * [JDK-8269096](https://bugs.openjdk.java.net/browse/JDK-8269096): Add java.util.Objects.newIdentity method

### Reviewers
 * [Mandy Chung](https://openjdk.java.net/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to bc73f6dd80dab95388baa7362285bc03706cab70
 * [Brian Burkhalter](https://openjdk.java.net/census#bpb) (@bplb - **Reviewer**) ⚠️ Review applies to bc73f6dd80dab95388baa7362285bc03706cab70
 * [Naoto Sato](https://openjdk.java.net/census#naoto) (@naotoj - **Reviewer**) ⚠️ Review applies to bc73f6dd80dab95388baa7362285bc03706cab70
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - **Reviewer**) ⚠️ Review applies to bc73f6dd80dab95388baa7362285bc03706cab70

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/112/head:pull/112` \
`$ git checkout pull/112`

`$ git checkout pull/112` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/112/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 112`

`$ git pr show -t 112`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/112.diff">https://git.openjdk.java.net/jdk17/pull/112.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/util/Objects.java
+++ b/src/java.base/share/classes/java/util/Objects.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2009, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -489,7 +489,7 @@ public final class Objects {
 
     /**
      * {@return a new instance of an unspecified class}
-     * The object has a unique identity&mdash; no other references to it exist.
+     * The object has a unique identity; no other references to it exist.
      * It can be used for synchronization, or where a placeholder Object is needed.
      * Use this method to avoid relying on the {@linkplain Object#Object() Object constructor}.
      *
--- a/src/java.base/share/classes/java/util/Objects.java
+++ b/src/java.base/share/classes/java/util/Objects.java
@@ -486,4 +486,16 @@ public final class Objects {
     long checkFromIndexSize(long fromIndex, long size, long length) {
         return Preconditions.checkFromIndexSize(fromIndex, size, length, null);
     }
+
+    /**
+     * {@return a new instance of an unspecified class}
+     * The object has a unique identity&mdash; no other references to it exist.
+     * It can be used for synchronization, or where a placeholder Object is needed.
+     * Use this method to avoid relying on the {@linkplain Object#Object() Object constructor}.
+     *
+     * @since 17
+     */
+    public static Object newIdentity() {
+        return new Object() {};
+    }
 }
--- a/src/java.base/share/classes/java/util/Objects.java
+++ b/src/java.base/share/classes/java/util/Objects.java
@@ -491,6 +491,7 @@ public final class Objects {
      * {@return a new instance of an unspecified class}
      * The object has a unique identity; no other references to it exist.
      * It can be used for synchronization, or where a placeholder Object is needed.
+     * The class does not override any of the methods of {@code java.lang.Object}.
      * Use this method to avoid relying on the {@linkplain Object#Object() Object constructor}.
      *
      * @since 17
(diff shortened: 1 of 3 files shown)


mlchung:
Looks okay in general.   It'd be useful for the javadoc of `Object()` constructor to suggest using `Objects::newIdentity` instead of the constructor.

mlchung (approved):

bplb (approved):

naotoj (approved):

src/java.base/share/classes/java/util/Objects.java
```diff
@@ -486,4 +486,16 @@ long checkFromToIndex(long fromIndex, long toIndex, long length) {
     long checkFromIndexSize(long fromIndex, long size, long length) {
         return Preconditions.checkFromIndexSize(fromIndex, size, length, null);
     }
+
+    /**
+     * {@return a new instance of an unspecified class}
+     * The object has a unique identity; no other references to it exist.
```
ChrisHegarty, line 492: Is this a new javadoc style/tag ?

RogerRiggs: Yes, it optimizes the case where the first sentence and the @ return tags have the same string.
See [JDK-8075778](https://bugs.openjdk.java.net/browse/JDK-8075778): Add javadoc tag to avoid duplication of return information in simple situations.

ChrisHegarty (approved):

test/jdk/java/util/Objects/BasicObjectsTest.java
```diff
@@ -45,6 +45,7 @@ public static void main(String... args) {
         errors += testIsNull();
         errors += testNonNull();
         errors += testNonNullOf();
+        errors += testNewIdentity();
```
JornVernee, line 48: The javadoc of `Objects::newIdentity` claims `[the returned object] can be used for synchronization`. Would it be useful to add a test that tries to synchronize on the result of `Objects.newIdentity` for that as well?

JornVernee (approved):

amaembo:
Probably it would be better to have an inner class named like `Identity` instead of anonymous class? When debugging or analyzing memory dumps, it would be more user-friendly to see `Objects$Identity` than `Objects$1`.

Probably, not the part of this feature request, but it would be nice to add another method with string parameter, like `Objects.newIdentity("MY SENTINEL")`. The string should be stored in the field and returned from toString(). Again, this would make it easier to find where the object comes from during debugging or memory dump analysis. For the record, here's what we have in IntelliJ IDEA sources (Apache 2.0 licensed):

```java
public final class ObjectUtils {
  private ObjectUtils() { }

  ...

  /**
   * Creates a new object which could be used as sentinel value (special value to distinguish from any other object). It does not equal
   * to any other object. Usually should be assigned to the static final field.
   *
   * @param name an object name, returned from {@link #toString()} to simplify the debugging or heap dump analysis
   *             (guaranteed to be stored as sentinel object field). If sentinel is assigned to the static final field,
   *             it's recommended to supply that field name (possibly qualified with the class name).
   * @return a new sentinel object
   */
  public static @NotNull Object sentinel(@NotNull @NonNls String name) {
    return new Sentinel(name);
  }

  private static final class Sentinel {
    private final String myName;

    Sentinel(@NotNull String name) {
      myName = name;
    }

    @Override
    public String toString() {
      return myName;
    }
  }
```

RogerRiggs:
@amaembo Good suggestion for the Valhalla API and implementation.  Today, only the hashCode() identifies an Object.
For JDK17, it will continue to be a raw Object() instance.

RogerRiggs:
The method proposed will be part of JEP 401.

PR #126, opened 2021-06-23
JDK-8269150 UnicodeReader not translating \u005c\\u005d to \\]

JimLaskey:
This issue relates to *Unicode escapes*, described in section 3.3 of the JLS. javac interprets Unicode escapes during the reading of ASCII characters from source. Later on, javac interprets *escape sequences*, described in section 3.7 of the JLS, during the tokenization of character literals, string literals, and text blocks. Escape sequences are only indirectly affected by this bug.

During reading, a _normal backslash_ (that is, the ASCII `\` character, not the corresponding Unicode escape `\u005c`) followed by another normal backslash is treated collectively as a pair of backslash characters. No further interpretation is done. This means that if a normal backslash immediately precedes the sequence `\` `u` `A` `B` `C` `D` which would "normally" be interpreted as an Unicode escape, then the interpretation of that sequence as a Unicode escape is suppressed.

For example, the sequence `\u2022` would be interpreted as the `•` character, whereas `\\u2022` would be interpreted as the seven characters `\` `\` `u` `2` `0` `2` `2`.

An issue arises when Java developers choose to use a _Unicode escape backslash_ `\u005c` in their source code, instead of a normal backslash. Prior to JDK 16, if the Unicode escape backslash was followed by a second Unicode escape, then *the second Unicode escape was always interpreted*. The normal backslash at the beginning of the second Unicode escape (immediately followed by `u`) was *not* paired with the preceding Unicode escape backslash. Elsewise, any following normal backslash will be paired with the `\u005c`.

For example, the sequence `\u005c\u2022` would be interpreted as `\` and `•`, whereas `\u005c\tXYZ` would be interpreted as `\` `\` `t` `X` `Y` `Z`.

The bug in JDK 16 ignored `\u005c` as having any effect on Unicode interpretation. Using the example from compiler-dev discussions, `\u005c\\u005d` :

- Prior to JDK 16, it was interpreted as `\` `\` `]`
- JDK 16 interpreted it as `\` `\` `\` `u` `0` `0` `5` `d` which would produce a syntax error downstream in the lexer because the escape sequence `\u` is invalid.

---------
### Progress

### Issue
 * [JDK-8269150](https://bugs.openjdk.java.net/browse/JDK-8269150): UnicodeReader not translating \u005c\\u005d to \\]

### Reviewers
 * [Jonathan Gibbons](https://openjdk.java.net/census#jjg) (@jonathan-gibbons - **Reviewer**) ⚠️ Review applies to 67efad992c03de1b7d3556e6b4df53f0ce2662b1
 * [Jan Lahoda](https://openjdk.java.net/census#jlahoda) (@lahodaj - **Reviewer**) ⚠️ Review applies to 67efad992c03de1b7d3556e6b4df53f0ce2662b1
 * [Joe Darcy](https://openjdk.java.net/census#darcy) (@jddarcy - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/126/head:pull/126` \
`$ git checkout pull/126`

`$ git checkout pull/126` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/126/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 126`

`$ git pr show -t 126`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/126.diff">https://git.openjdk.java.net/jdk17/pull/126.diff</a>

</details>


test/langtools/tools/javac/T8269150/T8269150.java
```diff
@@ -0,0 +1,36 @@
...
+/*
+ * @test
+ * @bug 8269150
+ * @summary Unicode \ u 0 0 5 C not treated as an escaping backslash
+ * @run compile T8269150.java
+ */
+
+public class T8269150 {
+    public static void main(String... args) {
+        System.err.println("\u005C\\u005D");
```
jonathan-gibbons, line 33: This seems to be a very weak "don't crash" test.
Shouldn't it check what is printed as well?

jonathan-gibbons (changes requested):

test/langtools/tools/javac/T8269150/T8269150.java
JimLaskey: Sure

JimLaskey:
I'll rework the fix to be consistent with jdk 15.

test/langtools/tools/javac/T8269150/T8269150.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+/*
+ * @test
+ * @bug 8269150
+ * @summary Unicode \ u 0 0 5 C not treated as an escaping backslash
+ * @run main T8269150
+ */
+
+public class T8269150 {
```
jddarcy, line 31: Please, please, please give the test a meaningful name other than a bug id. (It is true javac tested used to follow that naming convention, but it has largely been abandoned.)

test/langtools/tools/javac/T8269150/T8269150.java
```diff
@@ -0,0 +1,59 @@
...
+    public static void main(String... args) {
+        String source = """
```
jddarcy, line 33: The test here would be clearer if written as a 2-D array of strings where the alternative forms of a particular sequence of characters were side-by-side on the same line.

JimLaskey:
The fix has been revised to use the jdk15 (and before) logic. The JLS will be updated to clarify the fuzziness in this area.

lahodaj (approved):
Looks reasonable.

jonathan-gibbons (approved):

lahodaj (approved):
Looks good.

jddarcy:
Please revise the CSR for the updated change in behavior.

jddarcy:
@JimLaskey , please describe in text what the intended semantics are now for the escape processing. Thanks.

test/langtools/tools/javac/T8269150/T8269150.java
JimLaskey: Test expanded

test/langtools/tools/javac/T8269150/T8269150.java
JimLaskey: Done

test/langtools/tools/javac/T8269150/T8269150.java
JimLaskey: Done

jddarcy (approved):

PR #127, opened 2021-06-23
8268775: Password is being converted to String in AccessibleJPasswordField

azuev-java:
---------
### Progress

### Issue
 * [JDK-8268775](https://bugs.openjdk.java.net/browse/JDK-8268775): Password is being converted to String in AccessibleJPasswordField

### Reviewers
 * [Phil Race](https://openjdk.java.net/census#prr) (@prrace - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/127/head:pull/127` \
`$ git checkout pull/127`

`$ git checkout pull/127` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/127/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 127`

`$ git pr show -t 127`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/127.diff">https://git.openjdk.java.net/jdk17/pull/127.diff</a>

</details>

### DIFF
--- a/src/java.desktop/share/classes/javax/swing/JPasswordField.java
+++ b/src/java.desktop/share/classes/javax/swing/JPasswordField.java
@@ -515,11 +515,11 @@ public class JPasswordField extends JTextField {
             } else {
                 // Treat the text displayed in the JPasswordField
                 // as one word and sentence.
-                char[] password = getPassword();
-                if (password == null ||
-                    index < 0 || index >= password.length) {
+                int length = getDocument().getLength();
+                if (index < 0 || index >= length) {
                     return null;
                 }
+                char[] password = new char[length];
                 Arrays.fill(password, getEchoChar());
                 return new String(password);
             }
@@ -624,11 +624,11 @@ public class JPasswordField extends JTextField {
             } else {
                 // Treat the text displayed in the JPasswordField
                 // as one word, sentence, line and attribute run
-                char[] password = getPassword();
-                if (password == null ||
-                    index < 0 || index >= password.length) {
+                int length = getDocument().getLength();
+                if (index < 0 || index >= length) {
                     return null;
                 }
+                char[] password = new char[length];
                 Arrays.fill(password, getEchoChar());
                 String text = new String(password);
                 return new AccessibleTextSequence(0, password.length - 1, text);
--- a/src/java.desktop/share/classes/javax/swing/JPasswordField.java
+++ b/src/java.desktop/share/classes/javax/swing/JPasswordField.java
@@ -510,9 +510,8 @@ public class JPasswordField extends JTextField {
          * @since 1.6
          */
         public String getAtIndex(int part, int index) {
-           String str = null;
             if (part == AccessibleText.CHARACTER) {
-                str = super.getAtIndex(part, index);
+                return getEchoString(super.getAtIndex(part, index));
             } else {
                 // Treat the text displayed in the JPasswordField
                 // as one word and sentence.
@@ -521,9 +520,9 @@ public class JPasswordField extends JTextField {
                     index < 0 || index >= password.length) {
                     return null;
                 }
-                str = new String(password);
+                Arrays.fill(password, getEchoChar());
+                return new String(password);
             }
-            return getEchoString(str);
         }
 
         /**
@@ -544,8 +543,7 @@ public class JPasswordField extends JTextField {
          */
         public String getAfterIndex(int part, int index) {
             if (part == AccessibleText.CHARACTER) {
-                String str = super.getAfterIndex(part, index);
-                return getEchoString(str);
+                return getEchoString(super.getAfterIndex(part, index));
             } else {
                 // There is no word or sentence after the text
                 // displayed in the JPasswordField.
@@ -571,8 +569,7 @@ public class JPasswordField extends JTextField {
          */
         public String getBeforeIndex(int part, int index) {
             if (part == AccessibleText.CHARACTER) {
-                String str = super.getBeforeIndex(part, index);
-                return getEchoString(str);
+                return getEchoString(super.getBeforeIndex(part, index));
             } else {
                 // There is no word or sentence before the text
                 // displayed in the JPasswordField.
@@ -632,9 +629,9 @@ public class JPasswordField extends JTextField {
                     index < 0 || index >= password.length) {
                     return null;
                 }
+                Arrays.fill(password, getEchoChar());
                 String text = new String(password);
-                return new AccessibleTextSequence(0, password.length - 1,
-                                                  getEchoString(text));
+                return new AccessibleTextSequence(0, password.length - 1, text);
             }
         }
 


victordyakov:
@prrace @azvegint @prsadhuk  please review

prrace:
There's code but not a  single word of description of the fix. 
What is the problem, what are the changes here, why do they fix it, and what are the risks / consequences and why is there no test ? 
It is not very kind to reviewers to offer no explanation.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
```diff
@@ -510,9 +510,8 @@ private String getEchoString(String str) {
          * @since 1.6
          */
         public String getAtIndex(int part, int index) {
-           String str = null;
             if (part == AccessibleText.CHARACTER) {
-                str = super.getAtIndex(part, index);
+                return getEchoString(super.getAtIndex(part, index));
```
prrace, line 514: I don't see how removing the local variable changes anything. Explanation ??

src/java.desktop/share/classes/javax/swing/JPasswordField.java
```diff
@@ -544,8 +543,7 @@ public String getAtIndex(int part, int index) {
          */
         public String getAfterIndex(int part, int index) {
             if (part == AccessibleText.CHARACTER) {
-                String str = super.getAfterIndex(part, index);
-                return getEchoString(str);
+                return getEchoString(super.getAfterIndex(part, index));
```
prrace, line 546: I don't see how removing the local variable changes anything. Explanation ??

azuev-java:
The problem here is that if someone uses the accessibility methods on JPasswordField it will lead to unencrypted password being stored in the local String variable and that in turn can lead to it being recorded in, say, crash tump file where it can be found amongst the string literals. This is highly improbable scenario but it can be done so we better to not do it. And there are only two places where it happens - because in other methods we only serving the AccessibleText.CHARACTER retrieval which means that we are getting one password character at a time in a separate String variable which is Ok. These two places are at methods getAtIndex and getTextSequenceAt where we requesting something other than AccessibleText.CHARACTER. There we were converting password from the array of chars to the String only to pass this string to the method that generates string of echo characters of the same length. Instead i am doing conversion myself filling in the returned array with echo characters and returning constructed string. The rest of the changes is just a slight code cleanup - getting rid of the local variable that is used only to store some value before passing it to another method.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
azuev-java: Here it is just a slight code cleanup. We do not need additional variable for passing value from one method to another. It serves no other purpose at all.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
azuev-java: Here it is just a slight code cleanup. We do not need additional variable for passing value from one method to another. It serves no other purpose at all. It was used before on the second leg of the if but the usage was removed so it became useless.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
```diff
@@ -632,9 +629,9 @@ public AccessibleTextSequence getTextSequenceAt(int part, int index) {
                     index < 0 || index >= password.length) {
                     return null;
                 }
+                Arrays.fill(password, getEchoChar());
                 String text = new String(password);
-                return new AccessibleTextSequence(0, password.length - 1,
-                                                  getEchoString(text));
+                return new AccessibleTextSequence(0, password.length - 1, text);
```
prrace, line 634: So the accessible text is just the right number of "echo" chars.
And you are still calling getPassword() just so you can find out the length.
Then it is over-written. There's a really tiny window after getPassword() and before Arrays.fill() when the clear password is still there. 
The number of "char"s isn't the same as the number of "characters" if there's a non-BMP code point in there .. perhaps these are not allowed by this class .. but it makes me wonder how much if having the exact number of echo chars as the actual password is important.  I wonder how many text-to-speech readers can say "bullet" for a unicode bullet character ? 

If it weren't for all of this (the class and the getPassword() method being non-final I'd suggest you look into a way to pull just the length rather than the actual chars.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
prrace: Let me ask it this way.
Does super.getAfterIndex(part, index) return a String with any of the password in clear text ? 
It seems to me like it might.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
Vest: If this is about security, I don’t see how it might help. There is a chance that the heap dump might capture the content of the local variable. If you submit your heap dump to someone, whom you do not trust, I have bad news for you. 

The probability of the heap dump to capture a local variable is more than zero. True, but less than probable.

I don’t know if calling same methods in a single line makes this control more secure (if we take the situation that the heap dump pauses an execution of the thread exactly at our „moment of time“). I am not a member of the project JDK, but I doubt that this PR solves something. 

To me, an additional local variable adds better supportability (debugging) to this code. Otherwise everything should be put into a single fat method.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
azuev-java: For CHARACTER it will return String with a single character in the corresponding position. There is a possibility that someone will iterate the entirety of the password text and get all the characters in the password as a separate strings but digging it from the memory dump is much more difficult than the singular string with the whole password.
For anything but character we do not use this method - we get password as an array of characters and - after the fix - immediately overriding them with the same number of echo characters.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
azuev-java: > So the accessible text is just the right number of "echo" chars.
Yes.
> And you are still calling getPassword() just so you can find out the length.
> Then it is over-written. There's a really tiny window after getPassword() and before Arrays.fill() when the clear password is still there.

Yes. But not as a string - and the window of opportunity to get the characters before they are overwritten by the echo characters is minimal.

> The number of "char"s isn't the same as the number of "characters" if there's a non-BMP code point in there .. 

Since in order to enter non-BPM characters you have to have an input methods helper which should be disabled on password fields for obvious reason - it would pretty much disclose the typed password in the IM helper window

> perhaps these are not allowed by this class .. but it makes me wonder how much if having the exact number of echo chars as the actual password is important. I wonder how many text-to-speech readers can say "bullet" for a unicode bullet character ?
Well, accessibility is not only about text to speech - it is also about easier navigation so having exact number of the bullets is preferable. There are limitations - like some non-BPM text can be pasted into the password field and then navigating within it might be broken but since there will be no IM engaged fixing it would be equally problematic.

> If it weren't for all of this (the class and the getPassword() method being non-final I'd suggest you look into a way to pull just the length rather than the actual chars.

That would be preferable but under the current circumstances i would say that my fix makes things better without adding incompatible changes.

src/java.desktop/share/classes/javax/swing/JPasswordField.java
azuev-java: > I don’t know if calling same methods in a single line makes this control more secure

As i said above this exact change is not about making code more secure - it is just to eliminate additional variable that has no purpose after the second half of the method is changed. I would say that it would add to the supportability if we do anything with this information - but we don't.

victordyakov:
@mrserb  please review

src/java.desktop/share/classes/javax/swing/JPasswordField.java
mrserb: Why not use the "getDocument().getLength()"?

azuev-java: > Why not use the "getDocument().getLength()"?

We can surely do that. Fixed.

prrace (approved):

PR #130, opened 2021-06-23
8267952: async logging supports to dynamically change tags and decorators

navyxliu:
This patch fixed intermittent crashes of gtest/AsyncLogGtest.java. The root cause 
is that previous flush() can't guarantee flush all pending messages in AsyncLog
buffer. This patch implements flush() using a synchronizaton token and waits for
completion. This approach isn't MT-safe but it can serialize all flush() calls in
a thread. 

Two concurrent tests are appended to cover the hazard cases.
This patch also comments LogConfiguration::configure_output() MT-safety.

---------
### Progress

### Issue
 * [JDK-8267952](https://bugs.openjdk.java.net/browse/JDK-8267952): async logging supports to dynamically change tags and decorators

### Reviewers
 * [Paul Hohensee](https://openjdk.java.net/census#phh) (@phohensee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/130/head:pull/130` \
`$ git checkout pull/130`

`$ git checkout pull/130` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/130/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 130`

`$ git pr show -t 130`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/130.diff">https://git.openjdk.java.net/jdk17/pull/130.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/logging/logAsyncWriter.cpp
+++ b/src/hotspot/share/logging/logAsyncWriter.cpp
@@ -186,7 +186,7 @@ AsyncLogWriter* AsyncLogWriter::instance() {
 
 // Inserts a flush token into the async output buffer and waits until the AsyncLog thread
 // signals that it has seen it and completed all dequeued message processing.
-// This is method is not MT-safe in itself, but is guarded by another lock in the usual
+// This method is not MT-safe in itself, but is guarded by another lock in the usual
 // usecase - see the comments in the header file for more details.
 void AsyncLogWriter::flush() {
   if (_instance != nullptr) {
--- a/src/hotspot/share/logging/logAsyncWriter.cpp
+++ b/src/hotspot/share/logging/logAsyncWriter.cpp
@@ -41,7 +41,7 @@ class AsyncLogWriter::AsyncLogLocker : public StackObj {
 };
 
 void AsyncLogWriter::enqueue_locked(const AsyncLogMessage& msg) {
-  if (_buffer.size() >= _buffer_max_size)  {
+  if (_buffer.size() >= _buffer_max_size) {
     bool p_created;
     uint32_t* counter = _stats.add_if_absent(msg.output(), 0, &p_created);
     *counter = *counter + 1;
@@ -50,13 +50,12 @@ void AsyncLogWriter::enqueue_locked(const AsyncLogMessage& msg) {
     return;
   }
 
-  assert(_buffer.size() < _buffer_max_size, "_buffer is over-sized.");
   _buffer.push_back(msg);
   _sem.signal();
 }
 
 void AsyncLogWriter::enqueue(LogFileOutput& output, const LogDecorations& decorations, const char* msg) {
-  AsyncLogMessage m(output, decorations, os::strdup(msg));
+  AsyncLogMessage m(&output, decorations, os::strdup(msg));
 
   { // critical area
     AsyncLogLocker locker;
@@ -70,13 +69,13 @@ void AsyncLogWriter::enqueue(LogFileOutput& output, LogMessageBuffer::Iterator m
   AsyncLogLocker locker;
 
   for (; !msg_iterator.is_at_end(); msg_iterator++) {
-    AsyncLogMessage m(output, msg_iterator.decorations(), os::strdup(msg_iterator.message()));
+    AsyncLogMessage m(&output, msg_iterator.decorations(), os::strdup(msg_iterator.message()));
     enqueue_locked(m);
   }
 }
 
 AsyncLogWriter::AsyncLogWriter()
-  : _lock(1), _sem(0), _io_sem(1),
+  : _lock(1), _sem(0), _flush_sem(0),
     _initialized(false),
     _stats(17 /*table_size*/) {
   if (os::create_thread(this, os::asynclog_thread)) {
@@ -98,10 +97,10 @@ class AsyncLogMapIterator {
     using none = LogTagSetMapping<LogTag::__NO_TAG>;
 
     if (*counter > 0) {
-      LogDecorations decorations(LogLevel::Warning, none::tagset(), output->decorators());
+      LogDecorations decorations(LogLevel::Warning, none::tagset(), LogDecorators::All);
       stringStream ss;
       ss.print(UINT32_FORMAT_W(6) " messages dropped due to async logging", *counter);
-      AsyncLogMessage msg(*output, decorations, ss.as_string(true /*c_heap*/));
+      AsyncLogMessage msg(output, decorations, ss.as_string(true /*c_heap*/));
       _logs.push_back(msg);
       *counter = 0;
     }
@@ -118,7 +117,6 @@ void AsyncLogWriter::write() {
   // The operation 'pop_all()' is done in O(1). All I/O jobs are then performed without
   // lock protection. This guarantees I/O jobs don't block logsites.
   AsyncLogBuffer logs;
-  bool own_io = false;
 
   { // critical region
     AsyncLogLocker locker;
@@ -127,14 +125,11 @@ void AsyncLogWriter::write() {
     // append meta-messages of dropped counters
     AsyncLogMapIterator dropped_counters_iter(logs);
     _stats.iterate(&dropped_counters_iter);
-    own_io = _io_sem.trywait();
   }
 
   LinkedListIterator<AsyncLogMessage> it(logs.head());
-  if (!own_io) {
-    _io_sem.wait();
-  }
 
+  int req = 0;
   while (!it.is_empty()) {
     AsyncLogMessage* e = it.next();
     char* msg = e->message();
@@ -142,9 +137,17 @@ void AsyncLogWriter::write() {
     if (msg != nullptr) {
       e->output()->write_blocking(e->decorations(), msg);
       os::free(msg);
+    } else if (e->output() == nullptr) {
+      // encounter a flush token. take a record for the time being
+      // and notify flush() after the loop.
+      req++;
     }
   }
-  _io_sem.signal();
+
+  if (req > 0) {
+    assert(req == 1, "AsyncLogWriter::flush() is NOT MT-safe!");
+    _flush_sem.signal(req);
+  }
 }
 
 void AsyncLogWriter::run() {
@@ -181,10 +184,20 @@ AsyncLogWriter* AsyncLogWriter::instance() {
   return _instance;
 }
 
-// write() acquires and releases _io_sem even _buffer is empty.
-// This guarantees all logging I/O of dequeued messages are done when it returns.
+// NOT MT-safe! see the comments in the header file.
 void AsyncLogWriter::flush() {
   if (_instance != nullptr) {
-    _instance->write();
+    {
+      using none = LogTagSetMapping<LogTag::__NO_TAG>;
+      AsyncLogLocker locker;
+      LogDecorations d(LogLevel::Off, none::tagset(), LogDecorators::None);
+      AsyncLogMessage token(nullptr, d, nullptr);
+
+      // not disposable
+      _instance->_buffer.push_back(token);
+      _instance->_sem.signal();
+    }
+
+    _instance->_flush_sem.wait();
   }
 }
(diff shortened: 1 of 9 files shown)


src/hotspot/share/logging/logAsyncWriter.cpp
```diff
@@ -181,10 +184,20 @@ AsyncLogWriter* AsyncLogWriter::instance() {
   return _instance;
 }
 
-// write() acquires and releases _io_sem even _buffer is empty.
-// This guarantees all logging I/O of dequeued messages are done when it returns.
+// NOT MT-safe! see the comments in the header file.
```
dholmes-ora, line 187: Suggested comment:
// Inserts a flush token into the async output buffer and waits until the AsyncLog thread
// signals that it has seen it and completed all existing message processing.
// This is method is not MT-safe in itself, but is guarded by another lock in the usual
// usecase - see the comments in the header file for more details.

src/hotspot/share/logging/logAsyncWriter.cpp
```diff
@@ -181,10 +184,20 @@ AsyncLogWriter* AsyncLogWriter::instance() {
...
 void AsyncLogWriter::flush() {
   if (_instance != nullptr) {
-    _instance->write();
+    {
+      using none = LogTagSetMapping<LogTag::__NO_TAG>;
+      AsyncLogLocker locker;
+      LogDecorations d(LogLevel::Off, none::tagset(), LogDecorators::None);
+      AsyncLogMessage token(nullptr, d, nullptr);
+
+      // not disposable
```
dholmes-ora, line 196: I don't know what "not disposable" means.

src/hotspot/share/logging/logAsyncWriter.hpp
```diff
@@ -124,14 +124,13 @@ typedef KVHashtable<LogFileOutput*, uint32_t, mtLogging> AsyncLogMap;
 // instance() is MT-safe and returns the pointer of the singleton instance if and only if async logging is enabled and has well
```
dholmes-ora, line 124: s/well/successfully/

src/hotspot/share/logging/logAsyncWriter.hpp
```diff
@@ -124,14 +124,13 @@ typedef KVHashtable<LogFileOutput*, uint32_t, mtLogging> AsyncLogMap;
...
 // initialized. Clients can use its return value to determine async logging is established or not.
 //
-// The basic operation of AsyncLogWriter is enqueue(). 2 overloading versions of it are provided to match LogOutput::write().
+// enqueue() is the basic operation of AsyncLogWriter. 2 overloading versions of it are provided to match LogOutput::write().
```
dholmes-ora, line 127: s/2/Two/

src/hotspot/share/logging/logAsyncWriter.hpp
```diff
@@ -124,14 +124,13 @@ typedef KVHashtable<LogFileOutput*, uint32_t, mtLogging> AsyncLogMap;
...
 // They are both MT-safe and non-blocking. Derived classes of LogOutput can invoke the corresponding enqueue() in write() and
 // return 0. AsyncLogWriter is responsible of copying neccessary data.
 //
-// The static member function flush() is designated to flush out all pending messages when JVM is terminating.
-// In normal JVM termination, flush() is invoked in LogConfiguration::finalize(). flush() is MT-safe and can be invoked arbitrary
-// times. It is no-op if async logging is not established.
-//
+// flush() is designated to flush out all pending messages. MT-safety is not provided. It is no-op if async logging is not in use.
+// In normal JVM termination, flush() is invoked in LogConfiguration::finalize(). When the users change logging configurations
+// via jcmd, LogConfiguration::configure_output() invokes it with the protection of ConfigurationLock.
```
dholmes-ora, line 133: Suggestion:
// flush() ensures that all pending messages have been written out before it returns. It is not MT-safe
// in itself. When users change the logging configuration via jcmd, LogConfiguration::configure_output()
// calls flush() under the protection of the ConfigurationLock. In addition flush() is called during JVM
// termination, via LogConfiguration::finalize, but that occurs during the VM termination safepoint and so
// no calls to LogConfiguration::configure_output() can be active at the same time.

src/hotspot/share/logging/logAsyncWriter.hpp
```diff
@@ -141,8 +140,7 @@ class AsyncLogWriter : public NonJavaThread {
   // _sem is a semaphore whose value denotes how many messages have been enqueued.
   // It decreases in AsyncLogWriter::run()
   Semaphore _sem;
-  // A lock of IO
-  Semaphore _io_sem;
+  Semaphore _flush_sem;
```
dholmes-ora, line 143: Please add a comment:
// _flush_sem ensures flush() only returns after all enqueued messages are processed

src/hotspot/share/logging/logConfiguration.cpp
```diff
@@ -211,6 +211,21 @@ void LogConfiguration::delete_output(size_t idx) {
   delete output;
 }
 
+// MT-SAFETY
+//
+// ConfigurationLock can guarantee that only one thread is performing reconfiguration. This function still needs
```
dholmes-ora, line 216: // The ConfigurationLock guarantees ...

src/hotspot/share/logging/logConfiguration.cpp
```diff
@@ -211,6 +211,21 @@ void LogConfiguration::delete_output(size_t idx) {
...
+// to be MT-safe because logsites in other threads may be executing in parallel. Reconfiguration means unified
+// logging allows users to dynamically change tags and decorators of a log output via DCMD(logDiagnosticCommand.hpp).
+//
+// A RCU-style synchronization 'wait_until_no_readers()' is imposed inside of 'ts->set_output_level(output, level)'
```
dholmes-ora, line 220: s/imposed/used/

src/hotspot/share/logging/logConfiguration.cpp
```diff
@@ -211,6 +211,21 @@ void LogConfiguration::delete_output(size_t idx) {
...
+// if setting has changed. It guarantees that all logs either synchronous writing or enqueuing to the async buffer
```
dholmes-ora, line 221: // if a setting has changed. It guarantees that all logs, either synchronous writes or enqueuing to the async buffer,

src/hotspot/share/logging/logConfiguration.cpp
```diff
@@ -211,6 +211,21 @@ void LogConfiguration::delete_output(size_t idx) {
...
+// see the new tags and decorators. It's worth noting that the synchronization happens even level does not change.
```
dholmes-ora, line 222: s/even/even if the/

src/hotspot/share/logging/logConfiguration.cpp
```diff
@@ -253,6 +268,15 @@ void LogConfiguration::configure_output(size_t idx, const LogSelectionList& sele
     on_level[level]++;
   }
 
+  // for async logging, enqueuing instead of writing is still under protection of the synchronization
+  // `wait_until_no_readers()`. There are 2 hazards in async logging as follows.
+  //  1. asynclog buffer may be holding some log messages with previous decorators.
+  //  2. asynclog buffer may be holding some log messages targeting to the output 'idx'.
+  //
+  // A flush operation guarantees to all pending messages in buffer are written before returning. Therefore,
+  // the two hazards won't appear after it. It's a nop if async logging is not set.
```
dholmes-ora, line 277: Given all the other comments I think this can be simplified:

// For async logging we have to ensure that all enqueued messages, which may refer to previous decorators,
// or a soon-to-be-deleted outputs, are written out first. The flush() call ensures this. It is a no-op
// if async logging is not enabled.

dholmes-ora (changes requested):
Hi Xin,

Thank you for your patience and perseverance with this. 

This looks good to me now. The flush token seems to work well and is preferable to potentially blocking the async log thread.

I have a number of comments below, but they are all about comments.

I'm running this through our tier 1-7 testing again.

I have one remaining concern, which is just something we need to keep an eye on, and that is the potential for delaying VM termination if the async log writer gets blocked/delayed.

Thanks,
David
-----

src/hotspot/share/logging/logAsyncWriter.cpp
navyxliu: The reason I don't call enqueue_locked() here because it may be dropped due to buffer size. 
This is a control token. if it gets dropped, flush() will get stuck forever. 

not droppable?

dholmes-ora: How about:
// Push directly in-case we are at logical max capacity, as this must not get dropped

I'm curious now about the fact the buffer size seems only a logical size constraint not an actual out-of-space constraint.

navyxliu:
hi, David, 
> I have one remaining concern, which is just something we need to keep an eye on, and that is the potential for delaying VM termination if the async log writer gets blocked/delayed.

We discussed this before. Yes, if "write_blocking()" does block on I/O part, flush() is blocked.  We have this issue before, and synchronous logging has this issue as well.    I think Async logging is better than sync logging on this because at least JVM got stuck at termination instead of any safepoint if it did happen.  

It isn't perfect. Ideally, we can use flush_sem.wait(timeout) if Semaphore::wait supports timeout for all platforms. 

> I'm curious now about the fact the buffer size seems only a logical size constraint not an actual out-of-space constraint.

We can have an assertion like this. It's because flush() must be thread-safe. The reason I don't enforce it using mutex or yet another semaphore is that current use cases have already been protected. If all flush() are serialized, it means that there's one or none token at any time. therefore, the following assertion is right. Shall I put it back? 

`
assert(_buffer.size() < _buffer_max_size+1, "_buffer is over-sized.");
`

>  Warning: asynclog is OFF.
It is from gtest fixture AsyncLogTest, which is expected. We run those tests in two modes. 
Normally, gtest run this test in synchronous mode. It doesn't test much. 

https://github.com/openjdk/jdk/blob/master/test/hotspot/gtest/logging/test_asynclog.cpp#L37

Thank you for you patience. TBH, I coundn't wrap around my head if you did tell me that handshake problem. 
I will post a new revision to include your feedbacks.

thanks,
--lx

src/hotspot/share/logging/logAsyncWriter.cpp
navyxliu: currently, it's a soft limit.

src/hotspot/share/logging/logConfiguration.cpp
```diff
@@ -268,13 +268,8 @@ void LogConfiguration::configure_output(size_t idx, const LogSelectionList& sele
...
 
-  // for async logging, enqueuing instead of writing is still under protection of the synchronization
-  // `wait_until_no_readers()`. There are 2 hazards in async logging as follows.
-  //  1. asynclog buffer may be holding some log messages with previous decorators.
-  //  2. asynclog buffer may be holding some log messages targeting to the output 'idx'.
-  //
-  // A flush operation guarantees to all pending messages in buffer are written before returning. Therefore,
-  // the two hazards won't appear after it. It's a nop if async logging is not set.
+  // For async logging we have to ensure that all enqueued messages, which may refer to previous decorators,
+  // or a soon-to-be-deleted outputs, are written out first. The flush() call ensures this.
```
dholmes-ora, line 272: s/outputs/output/

src/hotspot/share/logging/logAsyncWriter.hpp
dholmes-ora: If you don't want to use my wording at least changed "designated" to "designed" as "designated is the wrong word. Thanks.

dholmes-ora (approved):
A couple of minor comment nits remain but otherwise this is fine.

Thanks,
David

navyxliu:
hi, @dcubed-ojdk 
Could you try to PR and see if it can solve the two crash sightings?

dholmes-ora (approved):

navyxliu:
Hi, Reviewers, 

May I have another reviewer to take a look at this?  This patch can fix the intermittent crashes for jdk17. 
By rules, we need to 2 reviewers to approve it before we can integrate it. David helped me verify it. 

thanks,
--lx

phohensee (approved):
Looks good.

PR #133, opened 2021-06-24
8269088: C2 fails with assert(!n->is_Store() && !n->is_LoadStore()) failed: no node with a side effect

rwestrel:
In PhaseIdealLoop::try_sink_out_of_loop(), a Store is not expected to
be a candidate for sinking: if a Store is in a loop, there has to be
at least a memory Phi keeping the Store in the loop. In the case of
the failure, a Store has control set in a loop but the Store is not
connected to a loop memory Phi. The reason for that is that the Store
was moved out of loop by PhaseIdealLoop::try_move_store_after_loop()
which moved the Store to the lca of its uses (other than the loop
Phi), in this case another loop.

It doesn't look right that a Store is removed from a loop only to be
added to another one so I propose fixing that by placing the Store
right out of the loop it's being moved out of (similar to
PhaseIdealLoop::try_sink_out_of_loop()) rather than at the lca of
uses.

---------
### Progress

### Issue
 * [JDK-8269088](https://bugs.openjdk.java.net/browse/JDK-8269088): C2 fails with assert(!n->is_Store() && !n->is_LoadStore()) failed: no node with a side effect

### Reviewers
 * [Vladimir Kozlov](https://openjdk.java.net/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to dcb1e6467a539aee184d751717a0a0bdf0f3a541
 * [Igor Veresov](https://openjdk.java.net/census#iveresov) (@veresov - **Reviewer**) ⚠️ Review applies to dcb1e6467a539aee184d751717a0a0bdf0f3a541
 * [Nils Eliasson](https://openjdk.java.net/census#neliasso) (@neliasso - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/133/head:pull/133` \
`$ git checkout pull/133`

`$ git checkout pull/133` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/133/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 133`

`$ git pr show -t 133`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/133.diff">https://git.openjdk.java.net/jdk17/pull/133.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/loopopts.cpp
+++ b/src/hotspot/share/opto/loopopts.cpp
@@ -950,7 +950,7 @@ void PhaseIdealLoop::try_move_store_after_loop(Node* n) {
             }
 #endif
             lca = place_outside_loop(lca, n_loop);
-            assert(!n_loop->is_member(get_loop(lca)), "control must not be back in of loop");
+            assert(!n_loop->is_member(get_loop(lca)), "control must not be back in the loop");
             assert(get_loop(lca)->_nest < n_loop->_nest || lca->in(0)->Opcode() == Op_NeverBranch, "must not be moved into inner loop");
 
             // Move store out of the loop
--- a/src/hotspot/share/opto/loopopts.cpp
+++ b/src/hotspot/share/opto/loopopts.cpp
@@ -949,6 +949,9 @@ void PhaseIdealLoop::try_move_store_after_loop(Node* n) {
               assert(get_loop(lca) == outer_loop, "safepoint in outer loop consume all memory state");
             }
 #endif
+            lca = place_outside_loop(lca, n_loop);
+            assert(!n_loop->is_member(get_loop(lca)), "control must not be back in of loop");
+            assert(get_loop(lca)->_nest < n_loop->_nest || lca->in(0)->Opcode() == Op_NeverBranch, "must not be moved into inner loop");
 
             // Move store out of the loop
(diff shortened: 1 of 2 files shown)


src/hotspot/share/opto/loopopts.cpp
```diff
@@ -949,6 +949,9 @@ void PhaseIdealLoop::try_move_store_after_loop(Node* n) {
               assert(get_loop(lca) == outer_loop, "safepoint in outer loop consume all memory state");
             }
 #endif
+            lca = place_outside_loop(lca, n_loop);
+            assert(!n_loop->is_member(get_loop(lca)), "control must not be back in of loop");
```
neliasso, line 953: assert comment needs adjusting

vnkozlov (approved):
@neliasso please run mach5 testing for this.

veresov (approved):

src/hotspot/share/opto/loopopts.cpp
rwestrel: Thanks for the review. Fixed.

neliasso (approved):

rwestrel:
@vnkozlov @veresov @neliasso thanks for the reviews

PR #143, opened 2021-06-24
8269335: Unable to load svml library

sviswa7:
The OpenJDK 17 early access build from jdk.java.net fails to load svml library. 
We need to give full svml library path to dll_load call for the load to succeed.,

---------
### Progress

### Issue
 * [JDK-8269335](https://bugs.openjdk.java.net/browse/JDK-8269335): Unable to load svml library

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/143/head:pull/143` \
`$ git checkout pull/143`

`$ git checkout pull/143` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/143/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 143`

`$ git pr show -t 143`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/143.diff">https://git.openjdk.java.net/jdk17/pull/143.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
+++ b/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
@@ -7001,17 +7001,9 @@ address generate_avx_ghash_processBlocks() {
     // Get svml stub routine addresses
     void *libsvml = NULL;
     char ebuf[1024];
-    char dll_path[JVM_MAXPATHLEN];
-#ifdef _WINDOWS
-    int ret = jio_snprintf(dll_path, sizeof(dll_path), "%s%sbin", Arguments::get_java_home(), os::file_separator());
-#else
-    int ret = jio_snprintf(dll_path, sizeof(dll_path), "%s%slib", Arguments::get_java_home(), os::file_separator());
-#endif
-    if (ret != -1) {
-      char dll_name[JVM_MAXPATHLEN];
-      if (os::dll_locate_lib(dll_name, sizeof(dll_name), dll_path, "svml")) {
-        libsvml = os::dll_load(dll_name, ebuf, sizeof ebuf);
-      }
+    char dll_name[JVM_MAXPATHLEN];
+    if (os::dll_locate_lib(dll_name, sizeof(dll_name), Arguments::get_dll_dir(), "svml")) {
+      libsvml = os::dll_load(dll_name, ebuf, sizeof ebuf);
     }
     if (libsvml != NULL) {
       // SVML method naming convention
--- a/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
+++ b/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
@@ -39,6 +39,7 @@
 #include "oops/objArrayKlass.hpp"
 #include "oops/oop.inline.hpp"
 #include "prims/methodHandles.hpp"
+#include "runtime/arguments.hpp"
 #include "runtime/frame.inline.hpp"
 #include "runtime/handles.inline.hpp"
 #include "runtime/sharedRuntime.hpp"
@@ -7000,7 +7001,12 @@ address generate_avx_ghash_processBlocks() {
     // Get svml stub routine addresses
     void *libsvml = NULL;
     char ebuf[1024];
-    libsvml = os::dll_load(JNI_LIB_PREFIX "svml" JNI_LIB_SUFFIX, ebuf, sizeof ebuf);
+    char dll_name[JVM_MAXPATHLEN];
+    int ret = jio_snprintf(dll_name, sizeof(dll_name), "%s%slib%s" JNI_LIB_PREFIX "svml" JNI_LIB_SUFFIX,
+                       Arguments::get_java_home(), os::file_separator(), os::file_separator());
+    if (ret != -1) {
+      libsvml = os::dll_load(dll_name, ebuf, sizeof ebuf);
+    }
     if (libsvml != NULL) {
       // SVML method naming convention
       //   All the methods are named as __svml_op<T><N>_ha_<VV>
--- a/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
+++ b/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
@@ -7001,11 +7001,13 @@ address generate_avx_ghash_processBlocks() {
     // Get svml stub routine addresses
     void *libsvml = NULL;
     char ebuf[1024];
-    char dll_name[JVM_MAXPATHLEN];
-    int ret = jio_snprintf(dll_name, sizeof(dll_name), "%s%slib%s" JNI_LIB_PREFIX "svml" JNI_LIB_SUFFIX,
-                       Arguments::get_java_home(), os::file_separator(), os::file_separator());
+    char dll_path[JVM_MAXPATHLEN];
+    int ret = jio_snprintf(dll_path, sizeof(dll_path), "%s%slib", Arguments::get_java_home(), os::file_separator());
     if (ret != -1) {
-      libsvml = os::dll_load(dll_name, ebuf, sizeof ebuf);
+      char dll_name[JVM_MAXPATHLEN];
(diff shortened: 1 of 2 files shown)


PaulSandoz (commented):
Unsure how this slipped through when we tested, but i think we need an explicit test. The test can execute a Java process with logging enabled and examine its log output (since the loading of SVML is logged). Searching the test code base should find some examples to copy.

DamonFool:
I used to test the svml in JDK17.
It seems to work on Linux.

So how can I reproduce this bug?
Thanks.

sviswa7:
@DamonFool The problem shows up only with the build on jdk.java.net. I also didn't see this problem while testing with internal builds.

DamonFool:
> @DamonFool The problem shows up only with the build on jdk.java.net. I also didn't see this problem while testing with internal builds.

So why it works for our internal builds but fails with the builds on jdk.java.net?

sviswa7:
@PaulSandoz I have attempted to use dll_locate_lib. Please see if this looks ok to you.

sviswa7:
@DamonFool Very good question. It took me sometime to figure this out. There is something in java.exe which makes the difference. If you replace the java.exe in the internal build with the one on jdk.java.net you will be able to reproduce the problem. Giving the full path to dll_load in libjvm.so as in this patch fixes the issue.

DamonFool:
> @DamonFool Very good question. It took me sometime to figure this out. There is something in java.exe which makes the difference. If you replace the java.exe in the internal build with the one on jdk.java.net you will be able to reproduce the problem. Giving the full path to dll_load in libjvm.so as in this patch fixes the issue.

I didn't get it tested on windows before.

Does the original code also work on Windows with your internal builds?
If so, I think more investigation is needed to figure out the root cause.
Thanks.

sviswa7:
@DamonFool The internal builds worked on both the platforms (Linux and Windows). Placing the built jvm.dll or libjvm.so with this patch in the jdk.java.net build also works on both the platforms.  
Please let me know if you have a better solution. This is the best that I could come up with. Please do keep in mind, we don't have too much time before RDP2.

PaulSandoz:
> @PaulSandoz I have attempted to use dll_locate_lib. Please see if this looks ok to you.

Be useful if someone from hotspot-runtime can look, @dholmes-ora WDYT?

We still need to add a test.

DamonFool:
> @DamonFool The internal builds worked on both the platforms (Linux and Windows). Placing the built jvm.dll or libjvm.so with this patch in the jdk.java.net build also works on both the platforms.
> Please let me know if you have a better solution. This is the best that I could come up with. Please do keep in mind, we don't have too much time before RDP2.

I'm not sure if the change really fix the problem since the root cause still remains unknown.

Replacing the jvm.dll or libjvm.so makes very little sense to me.

A better way to verify the fix is re-building the whole image with the patch on Oracle's platform.
But that is still not enough.
Even though it gets passed on Oracle's machines, we still don't know if it really get fixed on all others' platforms.

I don't have a Windows machine so it's hard for me to investigate it.
Thanks.

sviswa7:
@PaulSandoz I have added a test. Please let me know if it looks ok.

dholmes-ora:
@PaulSandoz I think we need input from the build team to determine why the java.net binary exhibits different behaviour to an internal build.

test/jdk/jdk/incubator/vector/LoadSvmlTest.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2020, 2021, Oracle and/or its affiliates. All rights reserved.
```
DamonFool, line 2: I think the copyright year should be 2021 only since it's a new file, right?

test/jdk/jdk/incubator/vector/LoadSvmlTest.java
```diff
@@ -0,0 +1,76 @@
...
+                av.lanewise(VectorOperators.SINH).intoArray(r, i);
+            }
+        }
+    }
+
+    public static void main(String... args) throws Exception {
+        ProcessBuilder pb = ProcessTools.createJavaProcessBuilder(
+            "-Xmn8m", "-Xlog:library=info",
+            "--add-modules=jdk.incubator.vector",
+            "LoadSvmlTest$VectorTest");
```
PaulSandoz, line 67: Can you do `VectorTest.class.getName()` instead?

PaulSandoz:
> @PaulSandoz I have added a test. Please let me know if it looks ok.

Test looks good. I will run through our build/test system tomorrow morning, and try to get some details on build differences.

test/jdk/jdk/incubator/vector/LoadSvmlTest.java
sviswa7: Thanks for catching it. I will correct the copyright year.

sviswa7:
Thanks a lot Paul.

src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
```diff
@@ -7000,7 +7001,18 @@ address generate_avx_ghash_processBlocks() {
     // Get svml stub routine addresses
     void *libsvml = NULL;
     char ebuf[1024];
-    libsvml = os::dll_load(JNI_LIB_PREFIX "svml" JNI_LIB_SUFFIX, ebuf, sizeof ebuf);
+    char dll_path[JVM_MAXPATHLEN];
+#ifdef _WINDOWS
+    int ret = jio_snprintf(dll_path, sizeof(dll_path), "%s%sbin", Arguments::get_java_home(), os::file_separator());
+#else
+    int ret = jio_snprintf(dll_path, sizeof(dll_path), "%s%slib", Arguments::get_java_home(), os::file_separator());
+#endif
```
dholmes-ora, line 7009: There is a platform independent way to do this. Here is the example of how libverify is loaded by the JVM:

// Load verify dll
  char buffer[JVM_MAXPATHLEN];
  char ebuf[1024];
  if (!os::dll_locate_lib(buffer, sizeof(buffer), Arguments::get_dll_dir(), "verify"))
    return NULL; // Caller will throw VerifyError

  void *lib_handle = os::dll_load(buffer, ebuf, sizeof(ebuf));

sviswa7:
@dholmes-ora Thanks a lot. I have updated the patch accordingly.

dholmes-ora:
I've checked the layout of the jdk.java.net binaries and those we use in our internal testing and see no difference, so I cannot explain why one case would fail when the other succeeds.

I'll take the latest patch for a spin through our build/test system and pass the link to @PaulSandoz to check progress/success.

David

test/jdk/jdk/incubator/vector/LoadSvmlTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        OutputAnalyzer output = new OutputAnalyzer(pb.start());
+        output.shouldHaveExitValue(0);
+        if (Platform.isWindows())
+          output.shouldMatch("Loaded library .*svml.dll");
+        else
+          output.shouldMatch("Loaded library .*libsvml.so");
```
dholmes-ora, line 74: You could just check for .*svml.* and not worry about dll vs so. That should suffice and keeps the test simpler.

sviswa7: Updated the test accordingly. Thanks.

PaulSandoz (approved):
All tests pass. I will talk with the build folks to find a root cause but i am confident it's more robust due to the use of `os::dll_locate_lib`.

If you can please update the test to use the class name, rather explicitly using the result of that.

sviswa7:
@PaulSandoz I have updated the test to use VectorTest.class.getName(). Please let me know if I can integrate.

PaulSandoz (approved):
Looks good.

PR #149, opened 2021-06-25
8269240: java/foreign/stackwalk/TestAsyncStackWalk.java test failed with concurrent GC

JornVernee:
This patch rewrites the prologue and epilogue of panama upcalls, in order to fix the test failure from the title.

Previously, we did a call to potentially attach the current thread to the VM, and then afterwards did the same suspend and stack reguard checks that we do on the back-edge of a native downcall. Then, on the back edge of the upcall we did another conditional call to detach the thread.

The suspend and reguard checks on the front-edge are incorrect, so I've changed the 2 calls to mimic what is done by JavaCallWrapper instead (with attach and detach included), and removed the old suspend and stack reguard checks.

FWIW, this removes the JavaFrameAnchor save/restore MacroAssembler code. This is now written in C++. Also, MacroAssembler code was added to save/restore the result of the upcall around the call on the back-edge, which was previously missing. Since the new code allocates a handle block as well, I've added handling for those oops to frame & OptimizedUpcallBlob.

Testing: local running of `jdk_foreign` on Windows and Linux (WSL). Tier 1-3

---------
### Progress

### Issue
 * [JDK-8269240](https://bugs.openjdk.java.net/browse/JDK-8269240): java/foreign/stackwalk/TestAsyncStackWalk.java test failed with concurrent GC

### Reviewers
 * [Vladimir Ivanov](https://openjdk.java.net/census#vlivanov) (@iwanowww - **Reviewer**) ⚠️ Review applies to c90416f52042813ef2c1a1c35d308fc631f77b2f

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/149/head:pull/149` \
`$ git checkout pull/149`

`$ git checkout pull/149` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/149/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 149`

`$ git pr show -t 149`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/149.diff">https://git.openjdk.java.net/jdk17/pull/149.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/frame_x86.cpp
+++ b/src/hotspot/cpu/x86/frame_x86.cpp
@@ -354,6 +354,7 @@ frame frame::sender_for_entry_frame(RegisterMap* map) const {
 }
 
 OptimizedEntryBlob::FrameData* OptimizedEntryBlob::frame_data_for_frame(const frame& frame) const {
+  assert(frame.is_optimized_entry_frame(), "wrong frame");
   // need unextended_sp here, since normal sp is wrong for interpreter callees
   return reinterpret_cast<OptimizedEntryBlob::FrameData*>(
     reinterpret_cast<char*>(frame.unextended_sp()) + in_bytes(_frame_data_offset));
--- a/src/hotspot/cpu/x86/frame_x86.cpp
+++ b/src/hotspot/cpu/x86/frame_x86.cpp
@@ -353,9 +353,10 @@ frame frame::sender_for_entry_frame(RegisterMap* map) const {
   return fr;
 }
 
-JavaFrameAnchor* OptimizedEntryBlob::jfa_for_frame(const frame& frame) const {
+OptimizedEntryBlob::FrameData* OptimizedEntryBlob::frame_data_for_frame(const frame& frame) const {
   // need unextended_sp here, since normal sp is wrong for interpreter callees
-  return reinterpret_cast<JavaFrameAnchor*>(reinterpret_cast<char*>(frame.unextended_sp()) + in_bytes(jfa_sp_offset()));
+  return reinterpret_cast<OptimizedEntryBlob::FrameData*>(
+    reinterpret_cast<char*>(frame.unextended_sp()) + in_bytes(_frame_data_offset));
 }
 
 bool frame::optimized_entry_frame_is_first() const {
--- a/src/hotspot/share/prims/universalUpcallHandler.cpp
+++ b/src/hotspot/share/prims/universalUpcallHandler.cpp
@@ -64,8 +64,6 @@ Thread* ProgrammableUpcallHandler::maybe_attach_and_get_thread(bool* should_deta
   } else {
     *should_detach = false;
   }
-  assert(thread->can_call_java(), "must be able to call Java");
-  guarantee(thread->as_Java_thread()->thread_state() == _thread_in_native, "must be comming from native");
   return thread;
 }
 
@@ -79,16 +77,18 @@ Thread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* conte
   JavaThread* thread = maybe_attach_and_get_thread(&context->should_detach)->as_Java_thread();
   context->thread = thread;
 
-  // Allocate handle block for Java code. This must be done before we change thread_state to _thread_in_Java_or_stub,
+  assert(thread->can_call_java(), "must be able to call Java");
+
+  // Allocate handle block for Java code. This must be done before we change thread_state to _thread_in_Java,
   // since it can potentially block.
   context->new_handles = JNIHandleBlock::allocate_block(thread);
 
-  // After this, we are official in JavaCode. This needs to be done before we change any of the thread local
+  // After this, we are official in Java Code. This needs to be done before we change any of the thread local
   // info, since we cannot find oops before the new information is set up completely.
   ThreadStateTransition::transition_from_native(thread, _thread_in_Java);
 
   // Make sure that we handle asynchronous stops and suspends _before_ we clear all thread state
-  // in JavaCallWrapper::JavaCallWrapper(). This way, we can decide if we need to do any pd actions
+  // in OptimizedEntryBlob::FrameData. This way, we can decide if we need to do any pd actions
   // to prepare for stop/suspend (flush register windows on sparcs, cache sp, or other state).
   bool clear_pending_exception = true;
   if (thread->has_special_runtime_exit_condition()) {
@@ -130,36 +130,29 @@ void ProgrammableUpcallHandler::on_exit(OptimizedEntryBlob::FrameData* context) 
 
   MACOS_AARCH64_ONLY(thread->enable_wx(WXWrite));
 
-  // restore previous handle block & Java frame linkage
+  // restore previous handle block
   thread->set_active_handles(context->old_handles);
 
   thread->frame_anchor()->zap();
 
   debug_only(thread->dec_java_call_counter());
 
-  // Old thread-local info. has been restored. We are not back in the VM.
+  // Old thread-local info. has been restored. We are not back in native code.
   ThreadStateTransition::transition_from_java(thread, _thread_in_native);
 
   // State has been restored now make the anchor frame visible for the profiler.
   // Do this after the transition because this allows us to put an assert
-  // the Java->vm transition which checks to see that stack is not walkable
+  // the Java->native transition which checks to see that stack is not walkable
   // on sparc/ia64 which will catch violations of the reseting of last_Java_frame
   // invariants (i.e. _flags always cleared on return to Java)
 
   thread->frame_anchor()->copy(&context->jfa);
 
-  // Release handles after we are marked as being inside the VM again, since this
+  // Release handles after we are marked as being in native code again, since this
   // operation might block
   JNIHandleBlock::release_block(context->new_handles, thread);
 
-  // I don't think this is really needed, but let's keep it for now
-  if (thread->has_pending_exception() && thread->has_last_Java_frame()) {
-    // If we get here, the Java code threw an exception that unwound a frame.
-    // It could be that the new frame anchor has not passed through the required
-    // StackWatermark barriers. Therefore, we process any such deferred unwind
-    // requests here.
-    StackWatermarkSet::after_unwind(thread);
-  }
+  assert(!thread->has_pending_exception(), "Upcall can not throw an exception");
 
   if (context->should_detach) {
     detach_thread(thread);
--- a/src/hotspot/share/prims/universalUpcallHandler.cpp
+++ b/src/hotspot/share/prims/universalUpcallHandler.cpp
@@ -51,7 +51,7 @@ void ProgrammableUpcallHandler::upcall_helper(JavaThread* thread, jobject rec, a
   JavaCalls::call_static(&result, upcall_method.klass, upcall_method.name, upcall_method.sig, &args, CATCH);
 }
 
-Thread* ProgrammableUpcallHandler::maybe_attach_and_get_thread(bool* should_detach) {
+JavaThread* ProgrammableUpcallHandler::maybe_attach_and_get_thread(bool* should_detach) {
   Thread* thread = Thread::current_or_null();
   if (thread == nullptr) {
     JavaVM_ *vm = (JavaVM *)(&main_vm);
@@ -64,7 +64,7 @@ Thread* ProgrammableUpcallHandler::maybe_attach_and_get_thread(bool* should_deta
   } else {
     *should_detach = false;
   }
-  return thread;
+  return thread->as_Java_thread();
 }
 
 void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
@@ -73,8 +73,8 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
 }
 
 // modelled after JavaCallWrapper::JavaCallWrapper
-Thread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* context) {
-  JavaThread* thread = maybe_attach_and_get_thread(&context->should_detach)->as_Java_thread();
+JavaThread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* context) {
+  JavaThread* thread = maybe_attach_and_get_thread(&context->should_detach);
   context->thread = thread;
 
   assert(thread->can_call_java(), "must be able to call Java");
@@ -83,13 +83,13 @@ Thread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* conte
   // since it can potentially block.
   context->new_handles = JNIHandleBlock::allocate_block(thread);
 
-  // After this, we are official in Java Code. This needs to be done before we change any of the thread local
+  // After this, we are officially in Java Code. This needs to be done before we change any of the thread local
   // info, since we cannot find oops before the new information is set up completely.
   ThreadStateTransition::transition_from_native(thread, _thread_in_Java);
 
   // Make sure that we handle asynchronous stops and suspends _before_ we clear all thread state
   // in OptimizedEntryBlob::FrameData. This way, we can decide if we need to do any pd actions
-  // to prepare for stop/suspend (flush register windows on sparcs, cache sp, or other state).
+  // to prepare for stop/suspend (cache sp, or other state).
   bool clear_pending_exception = true;
   if (thread->has_special_runtime_exit_condition()) {
     thread->handle_special_runtime_exit_condition();
@@ -111,10 +111,8 @@ Thread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* conte
   debug_only(thread->inc_java_call_counter());
   thread->set_active_handles(context->new_handles);     // install new handle block and reset Java frame linkage
 
-  assert (thread->thread_state() != _thread_in_native, "cannot set native pc to NULL");
-
   // clear any pending exception in thread (native calls start with no exception pending)
-  if(clear_pending_exception) {
+  if (clear_pending_exception) {
     thread->clear_pending_exception();
   }
 
@@ -140,12 +138,6 @@ void ProgrammableUpcallHandler::on_exit(OptimizedEntryBlob::FrameData* context) 
   // Old thread-local info. has been restored. We are not back in native code.
   ThreadStateTransition::transition_from_java(thread, _thread_in_native);
 
-  // State has been restored now make the anchor frame visible for the profiler.
-  // Do this after the transition because this allows us to put an assert
-  // the Java->native transition which checks to see that stack is not walkable
-  // on sparc/ia64 which will catch violations of the reseting of last_Java_frame
-  // invariants (i.e. _flags always cleared on return to Java)
-
   thread->frame_anchor()->copy(&context->jfa);
 
   // Release handles after we are marked as being in native code again, since this
@@ -161,7 +153,7 @@ void ProgrammableUpcallHandler::on_exit(OptimizedEntryBlob::FrameData* context) 
 
 void ProgrammableUpcallHandler::attach_thread_and_do_upcall(jobject rec, address buff) {
   bool should_detach = false;
-  Thread* thread = maybe_attach_and_get_thread(&should_detach);
+  JavaThread* thread = maybe_attach_and_get_thread(&should_detach);
 
   {
     MACOS_AARCH64_ONLY(ThreadWXEnable wx(WXWrite, thread));
--- a/src/hotspot/share/prims/universalUpcallHandler.cpp
+++ b/src/hotspot/share/prims/universalUpcallHandler.cpp
@@ -118,7 +118,7 @@ Thread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* conte
     thread->clear_pending_exception();
   }
 
-  MACOS_AARCH64_ONLY(_thread->enable_wx(WXExec));
+  MACOS_AARCH64_ONLY(thread->enable_wx(WXExec));
 
   return thread;
 }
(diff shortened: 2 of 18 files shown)


test/jdk/java/foreign/stackwalk/TestAsyncStackWalk.java
```diff
@@ -47,6 +47,34 @@
  *   TestAsyncStackWalk
  */
 
+/*
+ * @test id=zgc
+ * @requires (((os.arch == "amd64" | os.arch == "x86_64") & sun.arch.data.model == "64") | os.arch == "aarch64") & vm.gc.Z
```
JornVernee, line 52: I saw ZGC is not supported on some versions of Windows. Is this `vm.gc.Z` requires tag enough to exclude the test on those platforms?

test/jdk/java/foreign/stackwalk/TestStackWalk.java
```diff
@@ -47,6 +47,64 @@
...
+ *   -Xbootclasspath/a:.
+ *   -XX:+UnlockDiagnosticVMOptions
+ *   -XX:+WhiteBoxAPI
+ *   -Djdk.internal.foreign.ProgrammableInvoker.USE_INTRINSICS=false
+ *   --enable-native-access=ALL-UNNAMED
+ *   -Xbatch
+ *   -XX:+UseZGC
+ *   TestStackWalk
+ */
+
```
JornVernee, line 78:
```suggestion

```

test/jdk/java/foreign/stackwalk/TestAsyncStackWalk.java
```diff
@@ -47,6 +47,64 @@
...
+ *   -Xbootclasspath/a:.
+ *   -XX:+UnlockDiagnosticVMOptions
+ *   -XX:+WhiteBoxAPI
+ *   -Djdk.internal.foreign.ProgrammableInvoker.USE_INTRINSICS=false
+ *   --enable-native-access=ALL-UNNAMED
+ *   -Xbatch
+ *   -XX:+UseZGC
+ *   TestAsyncStackWalk
+ */
+
```
JornVernee, line 78:
```suggestion

```

src/hotspot/share/runtime/thread.hpp
```diff
@@ -1125,6 +1125,8 @@ class JavaThread: public Thread {
   void set_requires_cross_modify_fence(bool val) PRODUCT_RETURN NOT_PRODUCT({ _requires_cross_modify_fence = val; })
 
  private:
+  DEBUG_ONLY(void verify_frame_info();)
```
iwanowww, line 1128: If you declare `verify_frame_info` as returning a `bool` (and just put a `return true;` at the end of `JavaThread::verify_frame_info()`), you can call it as:
```
assert(verify_frame_info(), "unexpected frame info");
```

iwanowww (approved):
Looks good.

src/hotspot/cpu/x86/frame_x86.cpp
```diff
@@ -353,9 +353,10 @@ frame frame::sender_for_entry_frame(RegisterMap* map) const {
   return fr;
 }
 
-JavaFrameAnchor* OptimizedEntryBlob::jfa_for_frame(const frame& frame) const {
+OptimizedEntryBlob::FrameData* OptimizedEntryBlob::frame_data_for_frame(const frame& frame) const {
```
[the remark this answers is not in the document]
JornVernee: Good idea. Thanks

src/hotspot/share/runtime/thread.hpp
iwanowww: You can keep the assert inside `verify_frame_info()` which dumps additional data. (Yes, it's a bit confusing: assert inside an assert :-) )

iwanowww: FTR I'm fine with it either way.

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -71,6 +72,93 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
   vm->functions->DetachCurrentThread(vm);
 }
 
+// modelled after JavaCallWrapper::JavaCallWrapper
+Thread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* context) {
+  JavaThread* thread = maybe_attach_and_get_thread(&context->should_detach)->as_Java_thread();
```
dholmes-ora, line 77: Nit: if `maybe_attach_and_get_thread` has to return a JavaThread it should be typed to return a JavaThread.

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -71,6 +72,93 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
...
+  context->thread = thread;
+
+  assert(thread->can_call_java(), "must be able to call Java");
+
+  // Allocate handle block for Java code. This must be done before we change thread_state to _thread_in_Java,
+  // since it can potentially block.
+  context->new_handles = JNIHandleBlock::allocate_block(thread);
+
+  // After this, we are official in Java Code. This needs to be done before we change any of the thread local
```
dholmes-ora, line 86: typo: s/official/officially/
(I see this was copied from Javacalls)

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -71,6 +72,93 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
...
+  // info, since we cannot find oops before the new information is set up completely.
+  ThreadStateTransition::transition_from_native(thread, _thread_in_Java);
+
+  // Make sure that we handle asynchronous stops and suspends _before_ we clear all thread state
+  // in OptimizedEntryBlob::FrameData. This way, we can decide if we need to do any pd actions
+  // to prepare for stop/suspend (flush register windows on sparcs, cache sp, or other state).
```
dholmes-ora, line 92: No sparcs any more (again I see this was copied from Javacalls)

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -71,6 +72,93 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
...
+  // the valid transition is to clear _last_Java_sp and then reset the rest of
+  // the (platform specific) state.
+
+  context->jfa.copy(thread->frame_anchor());
+  thread->frame_anchor()->clear();
+
+  debug_only(thread->inc_java_call_counter());
+  thread->set_active_handles(context->new_handles);     // install new handle block and reset Java frame linkage
+
+  assert (thread->thread_state() != _thread_in_native, "cannot set native pc to NULL");
```
dholmes-ora, line 114: You transitioned to _thread_in_Java above so how can you possibly have changed state again ?? (okay again copied from javaCalls ...)

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -71,6 +72,93 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
...
+
+  // clear any pending exception in thread (native calls start with no exception pending)
+  if(clear_pending_exception) {
```
dholmes-ora, line 117: Nit: space after 'if'

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -71,6 +72,93 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
...
+    thread->clear_pending_exception();
+  }
+
+  MACOS_AARCH64_ONLY(thread->enable_wx(WXExec));
```
dholmes-ora, line 121: Not clear why this is needed? JavaCallWrapper doesn't use it.

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -71,6 +72,93 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
...
+
+  debug_only(thread->dec_java_call_counter());
+
+  // Old thread-local info. has been restored. We are not back in native code.
+  ThreadStateTransition::transition_from_java(thread, _thread_in_native);
+
+  // State has been restored now make the anchor frame visible for the profiler.
+  // Do this after the transition because this allows us to put an assert
+  // the Java->native transition which checks to see that stack is not walkable
+  // on sparc/ia64 which will catch violations of the reseting of last_Java_frame
```
dholmes-ora, line 146: Again archaic comment copied across :)

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -71,6 +72,93 @@ void ProgrammableUpcallHandler::detach_thread(Thread* thread) {
   vm->functions->DetachCurrentThread(vm);
 }
 
+// modelled after JavaCallWrapper::JavaCallWrapper
+Thread* ProgrammableUpcallHandler::on_entry(OptimizedEntryBlob::FrameData* context) {
```
dholmes-ora, line 76: This should return JavaThread not Thread.

src/hotspot/share/runtime/safepoint.cpp
```diff
@@ -928,6 +928,7 @@ void ThreadSafepointState::handle_polling_page_exception() {
   if( nm->is_at_poll_return(real_return_addr) ) {
     // See if return type is an oop.
     bool return_oop = nm->method()->is_returning_oop();
+    HandleMark hm(self);
```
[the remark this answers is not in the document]
dholmes-ora: How does native code reach a safepoint polling point?

src/hotspot/share/runtime/thread.cpp
```diff
@@ -1964,15 +1964,23 @@ void JavaThread::deoptimize_marked_methods() {
   }
 }
 
+#ifdef ASSERT
+void JavaThread::verify_frame_info() {
+  assert((!has_last_Java_frame() && java_call_counter() == 0) ||
+         (has_last_Java_frame() && java_call_counter() > 0),
+         "unexpected frame info: has_last_frame=%d, java_call_counter=%d",
+         has_last_Java_frame(), java_call_counter());
```
dholmes-ora, line 1972: I see this was relocated code, but as has_last_java_frame() returns bool, it shouldn't be treated as an int without a cast (or better use %s and report true or false).

dholmes-ora (commented):
Hi Jorn,

I can't comment on all the details here - especially in the x86_64 upcall handler code. The mapping to JavaCallWrapper seems reasonable but there are a few differences that I'm now assuming stem from the fact upcalls start _thread_in_native while JCW starts from _thread_in_vm?

Some minor comments and queries below (mostly issues with existing code that you copied).

Thanks,
David

src/hotspot/share/prims/universalUpcallHandler.cpp
JornVernee: Thanks, I wasn't sure whether to leave the comments in or not.

src/hotspot/share/prims/universalUpcallHandler.cpp
JornVernee: Yeah, it seemed paranoid to me as well. I'll remove this

src/hotspot/share/prims/universalUpcallHandler.cpp
JornVernee: I took this from JavaCallWrapper. Seems to have been added in 17 for the mach aarch64 port: https://github.com/openjdk/jdk17/blame/a32d2eefea12771522b942b32985df0fe50119e8/src/hotspot/share/runtime/javaCalls.cpp#L112

src/hotspot/share/runtime/safepoint.cpp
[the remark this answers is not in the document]
JornVernee: The stack trace looks like this:

```
Current thread (0x000002a2489f0b50):  JavaThread "Thread-4551" [_thread_in_Java, id=24920, stack(0x000000d9e0500000,0x000000d9e0600000)]

Stack: [0x000000d9e0500000,0x000000d9e0600000]
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [jvm.dll+0xae6651]  os::platform_print_native_stack+0xf1  (os_windows_x86.cpp:235)
V  [jvm.dll+0xda3f25]  VMError::report+0x1005  (vmError.cpp:739)
V  [jvm.dll+0xda58ae]  VMError::report_and_die+0x7fe  (vmError.cpp:1549)
V  [jvm.dll+0xda5fe4]  VMError::report_and_die+0x64  (vmError.cpp:1330)
V  [jvm.dll+0x4ceca7]  report_vm_error+0xb7  (debug.cpp:282)
V  [jvm.dll+0x6511be]  HandleArea::allocate_handle+0x3e  (handles.cpp:35)
V  [jvm.dll+0xb8e334]  ThreadSafepointState::handle_polling_page_exception+0x314  (safepoint.cpp:939)
C  0x000002a035d8caa7

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
v  ~SafepointBlob
J 639 c1 java.lang.invoke.LambdaForm$MH+0x000000080101b800.invoke(Ljava/lang/Object;)V java.base@18-internal (87 bytes) @ 0x000002a03635f7b4 [0x000002a03635f4a0+0x0000000000000314]
J 620 c1 java.lang.invoke.LambdaForm$MH+0x0000000801018c00.invoke(Ljava/lang/Object;)V java.base@18-internal (37 bytes) @ 0x000002a036353e0c [0x000002a036353720+0x00000000000006ec]
v  ~BufferBlob::mэYссссЫGў
```

So I think the 'native code' is something being called by the safepoint blob, but I'm not sure why it's marked with a `C` in the stack trace.

FWIW, this part has already been fixed as part of: https://github.com/openjdk/jdk17/pull/173 (not sure why it still shows up in the diff)

src/hotspot/share/prims/universalUpcallHandler.cpp
JornVernee: Thanks.

I've been careful here to return a `Thread*` since the result is stored in `r15_thread` and I thought converting between sub and super types could potentially result in different pointers due to the way super types are laid out within a subtype. I thought it worked like this:

```
Subclass
+---
| {Subclass vtable pointer}
| +--- (base class Super)
| | {Super vtable pointer}
| +---
+---
```

So, I thought plainly using a `JavaThread*` in generated machine code where a `Thread*` was expected could cause trouble, since the pointer needs to be offset for the type conversion.

But now that I'm looking at some cases with compiler explorer, the pointer offset only seems to be needed when using multiple inheritance, for instance:

```
class SuperA {
public:
    virtual void foo();
};

class SuperB {
public:
    virtual void bar();
};

class Sub : public SuperA, public SuperB {
public:
    virtual void baz();
};
```

Results in:

```
class Sub	size(16):
	+---
 0	| +--- (base class SuperA)
 0	| | {vfptr}
	| +---
 8	| +--- (base class SuperB)
 8	| | {vfptr}
	| +---
	+---

Sub::$vftable@SuperA@:
	| &Sub_meta
	|  0
 0	| &SuperA::foo 
 1	| &Sub::baz 

Sub::$vftable@SuperB@:
	| -8
 0	| &SuperB::bar 

Sub::baz this adjustor: 0
```

(https://godbolt.org/z/rq9bT8d9d)

It seems that the sub type just reuses the vtable pointer of the first super type (probably to avoid having to do this pointer offsetting). Though, converting between `SuperB*` and `Sub*` would require offsetting the pointer. I'm still not sure this is guaranteed to work like this with all compilers though (the example is with MSVC, which has options to dump class layouts).

The result of `on_entry` is stored in `r15_thread`, so I guess I'm wondering if it's safe to store a `JavaThread*` instead of a `Thread*` in `r15`, and other code, which may expect `r15` to hold a `Thread*`, is guaranteed to keep working? (FWIW, after changing the return type to `JavaThread*` the tests that exercise this code still pass on Windows with MSVC, and on WSL Linux with GCC).

src/hotspot/share/runtime/thread.cpp
JornVernee: > I can't comment on all the details here - especially in the x86_64 upcall handler code. The mapping to JavaCallWrapper seems reasonable but there are a few differences that I'm now assuming stem from the fact upcalls start _thread_in_native while JCW starts from _thread_in_vm?

JornVernee:
> The mapping to JavaCallWrapper seems reasonable but there are a few differences that I'm now assuming stem from the fact upcalls start _thread_in_native while JCW starts from _thread_in_vm?

The main difference stems from the fact that, since these upcalls can come from non-JNI native code, we can not assume that the thread is already attached to the VM, so we do that on the fly instead, and we detach the thread again after the upcall (if needed). Those are what the call to `maybe_attach_and_get_thread` at the start of `on_entry`, and `detach_thread` call at the end of `on_exit` are for.

The other thing is that there is no stack watermark check at the end of `on_exit`. This check is guarded by a check if the thread has any pending exceptions in `JavaCallWrapper`, but since a panama upcall is not allowed to throw any exceptions, I've changed that to an `assert` that checks that there are no pending exceptions at that point instead.

The last thing is that we transition directly from `_thread_in_native` to `_thread_in_Java`, which changes some of the thread transition code.

Other than that, I've tried to mimic what `JavaCallWrapper` does as closely as possible.

Is there anything else that looks different?

src/hotspot/share/prims/universalUpcallHandler.cpp
JornVernee: Sorry, I sent the wrong godbolt link: https://godbolt.org/z/1665fWzff

JornVernee:
David, I've addressed your review comments.

For now I went with changing the return type of `on_entry` to `JavaThread*`, assuming the potential pointer conversion is not an issue. 

I tried for a while to implement a static assert to check that `Thread*` is trivial convertible to `JavaThread*` as well, but couldn't find a way to do that at compile time, so left it for now. It looks like C++ 20 has a type trait to check this: https://en.cppreference.com/w/cpp/types/is_pointer_interconvertible_base_of (I think this does what we want). I thought maybe we could just mimic the reference implementation, but looking at for instance the MSVC STL implementation, this type trait is implemented as a compiler intrinsic, so I think we'll have to wait until we are on C++ 20 before adding such a static assert.

src/hotspot/share/runtime/safepoint.cpp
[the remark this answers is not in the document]
dholmes-ora: Okay so we're not actually _thread_in_native it is just a chunk of VM generated code. Something still seems off to me about the need for the HandleMark but that isn't your problem.

src/hotspot/share/prims/universalUpcallHandler.cpp
vnkozlov: @JornVernee I have small correct to your comment. We use simple inheritance for Thread subclasses. Their instances have **one** vtbl pointer at the same offset. But this pointer will point to separate vtable for each subclass (and base class). The layout (sequence) of methods pointers in vtable is the same in base class and subclasses. But subclass specific methods pointers will be different.

The only issue is that you have to make sure to cast passed object pointer to correct subclass (or base class). Otherwise you will get incorrect vtable and incorrect virtual methods pointers.

R15 is used by our JIT compiled code and Interpreter code which are executed only in JavaThread so the pinter it contains is JavaThread*

src/hotspot/share/prims/universalUpcallHandler.cpp
dholmes-ora: Sorry not sure what I was looking at when I didn't see this.

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -60,20 +60,100 @@ Thread* ProgrammableUpcallHandler::maybe_attach_and_get_thread(bool* should_deta
     guarantee(result == JNI_OK, "Could not attach thread for upcall. JNI error code: %d", result);
     *should_detach = true;
     thread = Thread::current();
```
dholmes-ora, line 62: You could use JavaThread::current() here and avoid the later conversions.

src/hotspot/share/prims/universalUpcallHandler.cpp
```diff
@@ -140,12 +138,6 @@ void ProgrammableUpcallHandler::on_exit(OptimizedEntryBlob::FrameData* context)
   // Old thread-local info. has been restored. We are not back in native code.
```
dholmes-ora, line 138: Pre-existing: I think "not" should be "now".

dholmes-ora (approved):
Thanks for the updates.

Minor comments below.

David

src/hotspot/share/prims/universalUpcallHandler.cpp
JornVernee: Ok. The variable is initially initialized (hah) using `Thread::current_or_null()` and there was no equivalent that does that for `JavaThread`. I can add something like that though.

src/hotspot/share/prims/universalUpcallHandler.cpp
JornVernee: Thanks Vladimir. This also matches what I was seeing in compiler explorer, but I wasn't sure if we could assume it always worked like that with every C++ compiler.

It sound like R15 is expected to hold a `JavaThread*` though, so making the return type of `on_entry` be `JavaThread*` as David suggested seems correct.

Thanks

JornVernee:
I've addressed the new review comments, changing some `Thread*` to `JavaThread*`, removing now redundant casts, and finally noticed that the `Thread*` being passed to `ProgrammableUpcallHandle::detach_thread` was not being used, and it was always detaching the current thread instead. This in itself is fine, but then there's no need for the parameter, so I've dropped that.

src/hotspot/share/runtime/thread.hpp
```diff
@@ -1423,6 +1423,8 @@ class JavaThread: public Thread {
  public:
   // Returns the running thread as a JavaThread
   static inline JavaThread* current();
+  // Returns the current thread as a JavaThread, or NULL if not attached
+  static inline JavaThread* current_or_null();
```
dholmes-ora, line 1427: I hadn't intended to suggest the introduction of this method, but I see now how my comment led to it. It is fine.

dholmes-ora (approved):
Updates look good.

Thanks,
David

JornVernee:
> I can't comment on all the details here - especially in the x86_64 upcall handler code.

Vladimir's review covers that part of the patch (confirmed with him), so I will go ahead and integrate this.

PR #156, opened 2021-06-28
8269230: C2: main loop in micro benchmark never executed

rwestrel:
The benchmark loop is roughly:

for (int i = 0; i < limit; i++) {
  if (!(i < max_jint && i > min_jint)) {
    uncommon_trap();
  }
}

which IfNode::fold_compares() tranforms to:

for (int i = 0; i < limit; i++) {
  if (!(i - min_jint - 1 <u -2)) {
    uncommon_trap();
  }
}

Range elimination then adjusts loop limits so the unsigned comparison
can be removed. The new limit is:

min(limit, max(-2 + min_jint + 1, min_jint)) = min(limit, min_jint) = min_jint

and the main loop is never executed.

The logic that protects against overflows during range check
elimination is what causes the main loop to never be executed. As a
fix, I propose delaying IfNode::fold_compares() if the resulting test
compares with a potential negative number which could then cause RC
overflow code to trigger. I think that would preserve the most
interesting cases for IfNode::fold_compares() when it comes to range
check elimination:

i >= 0 && i < array.length

---------
### Progress

### Issue
 * [JDK-8269230](https://bugs.openjdk.java.net/browse/JDK-8269230): C2: main loop in micro benchmark never executed

### Reviewers
 * [Igor Veresov](https://openjdk.java.net/census#iveresov) (@veresov - **Reviewer**)

### Contributors
 * Maurizio Cimadamore `<mcimadamore@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/156/head:pull/156` \
`$ git checkout pull/156`

`$ git checkout pull/156` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/156/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 156`

`$ git pr show -t 156`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/156.diff">https://git.openjdk.java.net/jdk17/pull/156.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/ifnode.cpp
+++ b/src/hotspot/share/opto/ifnode.cpp
@@ -1011,6 +1011,33 @@ bool IfNode::fold_compares_helper(ProjNode* proj, ProjNode* success, ProjNode* f
       adjusted_lim = igvn->transform(new SubINode(hi, lo));
     }
     hook->destruct(igvn);
+
+    int lo = igvn->type(adjusted_lim)->is_int()->_lo;
+    if (lo < 0) {
+      // If range check elimination applies to this comparison, it includes code to protect from overflows that may
+      // cause the main loop to be skipped entirely. Delay this transformation.
+      // Example:
+      // for (int i = 0; i < limit; i++) {
+      //   if (i < max_jint && i > min_jint) {...
+      // }
+      // Comparisons folded as:
+      // i - min_jint - 1 <u -2
+      // when RC applies, main loop limit becomes:
+      // min(limit, max(-2 + min_jint + 1, min_jint))
+      // = min(limit, min_jint)
+      // = min_jint
+      if (!igvn->C->post_loop_opts_phase()) {
+        if (adjusted_val->outcnt() == 0) {
+          igvn->remove_dead_node(adjusted_val);
+        }
+        if (adjusted_lim->outcnt() == 0) {
+          igvn->remove_dead_node(adjusted_lim);
+        }
+        igvn->C->record_for_post_loop_opts_igvn(this);
+        return false;
+      }
+    }
+
     Node* newcmp = igvn->transform(new CmpUNode(adjusted_val, adjusted_lim));
     Node* newbool = igvn->transform(new BoolNode(newcmp, cond));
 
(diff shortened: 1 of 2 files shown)


vnkozlov:
I have to look but I think RCE mix up unsigned and signed expressions when calculating new limit. `min(limit, min_jint)` should be unsigned I think.

vnkozlov (commented):
I have to look but I think RCE mix up unsigned and signed expressions when calculating new limit. `min(limit, min_jint)` should be unsigned I think.

dean-long:
What does the code tranform to after range check elimination?

rwestrel:
> I have to look but I think RCE mix up unsigned and signed expressions when calculating new limit. `min(limit, min_jint)` should be unsigned I think.

FYI, https://github.com/openjdk/jdk/pull/4362 changes that code but doesn't fix that issue.

rwestrel:
> What does the code transform to after range check elimination?

The main loop body has no test in the final code (with the proposed fix) and performance is on par with other micro benchmarks that this benchmark is compared with.

RCE has logic to handle signed comparisons so I expect that RCE is what removes the i < max_jint and i > min_jint from the main loop body.

vnkozlov (changes requested):
I can't approve this for JDK 17. The issue is performance for very corner case but suggested fix may affect other cases. And this is not a bug IMHO - it is RFE which we should spend more time to investigate. I would suggest to defer it to JDK 18.

vnkozlov:
Okay. Maurizio and Paul explained to me in the bug report that it is very important code pattern in foreign memory API because C2 still does not support RCE over long index loops [8259609](https://bugs.openjdk.java.net/browse/JDK-8259609).
Taking this into account I think it is fine to push it into JDK 17 after testing which I will do.

vnkozlov (commented):
I will test it and approve after that if everything (including performance) is good.

vnkozlov:
Add the test to test/micro/org/openjdk/bench/jdk/incubator/foreign/

vnkozlov:
tier1-3 testing passed. Performance results will be tomorrow.

rwestrel:
> Add the test to test/micro/org/openjdk/bench/jdk/incubator/foreign/

done

mcimadamore:
I tried this patch on my machine - I can confirm it solves the problem, and that the benchmark shows similar numbers in all iterations now. Thanks for taking care of this!

rwestrel:
> > Add the test to test/micro/org/openjdk/bench/jdk/incubator/foreign/
> 
> done

And I added Maurizio as contributor of this change so he gets credit for the test case.

vnkozlov (approved):
Performance results are good - no regression.
You need second review.

veresov (approved):
Looks reasonable

rwestrel:
@vnkozlov @veresov thanks for the reviews

PR #162, opened 2021-06-28
JDK-8269426:  Move tests from test/jdk/java/lang/invoke/t8150782 to its parent directory

mlchung:
`test/jdk/java/lang/invoke/t8150782` is the only one using that convention. There are several tests under `<bugid>` directory.  As the tests under `test/jdk/java/lang/invoke/t8150782` are mostly unit tests for `Lookup::accessClass` and `Lookup::findClass`, they can simply be moved to the parent directory.


JornVernee:
Hi Mandy,

It looks like the test in question has some test support files associated with it (p/Foo.java & q/Bar.java). I personally think it's nice if tests like these are in a separate directory, so that it's easier to see which test support files belong to which test (and to avoid potential name clashes when 2 tests want to use a support file with the same name).

How would you feel about dropping just the `t` prefix from the test directory & package names? Then it would match the existing tests in `<bugid>` directories.

mlchung:
That's a fair point.  I personally prefer using a descriptive directory name rather than bug ID since that can give the reader what the tests are intended for.   What about renaming `t8150782` to `accessClassAndFindClass`?

JornVernee:
> That's a fair point. I personally prefer using a descriptive directory name rather than bug ID since that can give the reader what the tests are intended for. What about renaming t8150782 to accessClassAndFindClass?

Me too! That sounds like an excellent idea :)

JornVernee (approved):

mlchung:
Thanks Jorn.

PR #166, opened 2021-06-28
8269543: The warning for System::setSecurityManager should only appear once for each caller

wangweij:
Add a cache to record which sources have called `System::setSecurityManager` and only print out warning lines once for each.

---------
### Progress

### Issue
 * [JDK-8269543](https://bugs.openjdk.java.net/browse/JDK-8269543): The warning for System::setSecurityManager should only appear once for each caller

### Reviewers
 * [Lance Andersen](https://openjdk.java.net/census#lancea) (@LanceAndersen - **Reviewer**)
 * [Alan Bateman](https://openjdk.java.net/census#alanb) (@AlanBateman - **Reviewer**)
 * [Daniel Fuchs](https://openjdk.java.net/census#dfuchs) (@dfuch - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/166/head:pull/166` \
`$ git checkout pull/166`

`$ git checkout pull/166` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/166/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 166`

`$ git pr show -t 166`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/166.diff">https://git.openjdk.java.net/jdk17/pull/166.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/System.java
+++ b/src/java.base/share/classes/java/lang/System.java
@@ -331,7 +331,7 @@ public final class System {
     private static class CallersHolder {
         // Remember callers of setSecurityManager() here so that warning
         // is only printed once for each different caller
-        final static Map<String, Boolean> callers
+        final static Map<Class<?>, Boolean> callers
             = Collections.synchronizedMap(new WeakHashMap<>());
     }
 
@@ -387,14 +387,14 @@ public final class System {
     public static void setSecurityManager(@SuppressWarnings("removal") SecurityManager sm) {
         if (allowSecurityManager()) {
             var callerClass = Reflection.getCallerClass();
-            URL url = codeSource(callerClass);
-            final String source;
-            if (url == null) {
-                source = callerClass.getName();
-            } else {
-                source = callerClass.getName() + " (" + url + ")";
-            }
-            if (CallersHolder.callers.putIfAbsent(source, true) == null) {
+            if (CallersHolder.callers.putIfAbsent(callerClass, true) == null) {
+                URL url = codeSource(callerClass);
+                final String source;
+                if (url == null) {
+                    source = callerClass.getName();
+                } else {
+                    source = callerClass.getName() + " (" + url + ")";
+                }
                 initialErrStream.printf("""
                         WARNING: A terminally deprecated method in java.lang.System has been called
                         WARNING: System::setSecurityManager has been called by %s
--- a/src/java.base/share/classes/java/lang/System.java
+++ b/src/java.base/share/classes/java/lang/System.java
@@ -328,10 +328,10 @@ public final class System {
     private static native void setOut0(PrintStream out);
     private static native void setErr0(PrintStream err);
 
-    private static class CallerHolder {
+    private static class CallersHolder {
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/java/lang/System.java
```diff
@@ -326,6 +328,13 @@ private static void checkIO() {
     private static native void setOut0(PrintStream out);
     private static native void setErr0(PrintStream err);
 
+    private static class CallersHolder {
+        // Remember callers of setSecurityManager() here so that warning
+        // is only printed once for each different caller
+        final static Map<String, Boolean> callers
+            = Collections.synchronizedMap(new WeakHashMap<>());
+    }
+
```
dfuch, line 337: I wonder about the use of a WeakHashMap here. That may work well when the source is an interned string (a class name) which will be strongly referenced elsewhere and may be garbage collected if the class is unloaded, but in the case where it contains the name of the source jar then that string will only be referenced by the weak hashmap, and therefore it could be garbage collected any time - which would cause the mapping to be removed. In that case you cannot guarantee that the warning will be emitted only once.

RogerRiggs: Using a HashSet<Class> could use the callerClass as the key and be a stable reference for having given the message.
or use a ConcurrentHashMap<Class<?>>, boolean> and avoid any separate synchronization that would be needed with a HashSet or HashMap.

wangweij: If I switch to a "non-weak" set or map, then it seems I can safely use the source string as the key. Will using the Class object as a key prevent them from unloading?

RogerRiggs: Using a synchronized WeakHashMap with the class as the key would not prevent class unloading.
Using a non-weak set or map to strings would keep the strings around for the life of the runtime.

wangweij: I hope this is uncommon but if that class is created by a `ClassLoader` again and again then it will be difference each time. I'll investigate more.

AlanBateman: WeakHashMap<Class<?>, Boolean>, where the key is the caller, should be okay (assume synchronization of course). Even with applications that do call setSecurityManager then the map is probably going to be one entry. I wouldn't expect it is common to create custom class loaders that load code that sets a security manager, meaning it is more likely that the container sets the SM rather have each plugin/application/whatever try to set the system-wide SM.

wangweij: Thanks. Switched to Class<?> as cache key. New commit pushed.

jaikiran:
FWIW - I built a local JDK with this PR and gave it a try against one of the original Apache Ant project example where the warnings were previously flooding the stderr. With this change, I now see the warning message just once (as expected) for that example.

LanceAndersen (approved):

AlanBateman (approved):

dfuch (approved):

PR #168, opened 2021-06-29
8269568: JVM crashes when running VectorMask query tests

XiaohongGong:
This is a follow-up patch for [1]. When we are trying to add the VectorMask query implementation for Arm NEON, we found the jtreg tests for `VectorMask.trueCount/firstTrue/lastTrue` is not effective. The tests failure cannot be reported as expected. The main reason is that the Vector API methods are not hot enough to be compiled by C2 compiler. Wrap the main test codes inside a loop can make the tests effective.

With the tests taking effect, we can see a JVM crash due to the following assertion:
```
  vector/src/hotspot/share/opto/vectornode.hpp:879), pid=168241, tid=168257
  # Error: assert(mask->Opcode() == Op_VectorStoreMask) failed
```
The mask input might be other vector nodes like `"LoadVectorNode"`, since there is an optimization for `"VectorStoreMask"`:
```
  VectorStoreMask (VectorLoadMask value) ==> value
```
Changing the code to check whether its element basic type is `"T_BOOLEAN"` is more reasonable.

[1] https://bugs.openjdk.java.net/browse/JDK-8256973

---------
### Progress

### Issue
 * [JDK-8269568](https://bugs.openjdk.java.net/browse/JDK-8269568): JVM crashes when running VectorMask query tests

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - **Reviewer**) ⚠️ Review applies to 8e3fcfaef51b2f555fec4bbe7eb53412696f8555
 * [Jie Fu](https://openjdk.java.net/census#jiefu) (@DamonFool - **Reviewer**) ⚠️ Review applies to 8e3fcfaef51b2f555fec4bbe7eb53412696f8555
 * [Jatin Bhateja](https://openjdk.java.net/census#jbhateja) (@jatin-bhateja - Committer)
 * [Sandhya Viswanathan](https://openjdk.java.net/census#sviswanathan) (@sviswa7 - **Reviewer**)

### Contributors
 * Sandhya Viswanathan `<sviswanathan@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/168/head:pull/168` \
`$ git checkout pull/168`

`$ git checkout pull/168` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/168/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 168`

`$ git pr show -t 168`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/168.diff">https://git.openjdk.java.net/jdk17/pull/168.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -3863,6 +3863,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vpsubb(xtmp, xtmp, mask, vec_enc);
   evpmovb2m(ktmp, xtmp, vec_enc);
   kmovql(tmp, ktmp);
+  if (masklen < 64) {
+    andq(tmp, (((jlong)1 << masklen) - 1));
+  }
   switch(opc) {
     case Op_VectorMaskTrueCount:
       popcntq(dst, tmp);
@@ -3887,6 +3890,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vpxor(xtmp, xtmp, xtmp, vec_enc);
   vpsubb(xtmp, xtmp, mask, vec_enc);
   vpmovmskb(tmp, xtmp, vec_enc);
+  if (masklen < 64) {
+    andq(tmp, (((jlong)1 << masklen) - 1));
+  }
   switch(opc) {
     case Op_VectorMaskTrueCount:
       popcntq(dst, tmp);
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -3863,9 +3863,6 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vpsubb(xtmp, xtmp, mask, vec_enc);
   evpmovb2m(ktmp, xtmp, vec_enc);
   kmovql(tmp, ktmp);
-  if (masklen < 64) {
-    andq(tmp, (((jlong)1 << masklen) - 1));
-  }
   switch(opc) {
     case Op_VectorMaskTrueCount:
       popcntq(dst, tmp);
--- a/test/jdk/jdk/incubator/vector/templates/Unit-Miscellaneous.template
+++ b/test/jdk/jdk/incubator/vector/templates/Unit-Miscellaneous.template
@@ -415,14 +415,16 @@
     static void maskTrueCount$vectorteststype$SmokeTest(IntFunction<boolean[]> fa) {
         boolean[] a = fa.apply(SPECIES.length());
 
-        for (int i = 0; i < a.length; i += SPECIES.length()) {
-            var vmask = SPECIES.loadMask(a, i);
-            int tcount = vmask.trueCount();
-            int expectedTcount = 0;
-            for (int j = i; j < i + SPECIES.length(); j++) {
-                expectedTcount += a[j] ? 1 : 0;
+        for (int ic = 0; ic < INVOC_COUNT * INVOC_COUNT; ic++) {
+            for (int i = 0; i < a.length; i += SPECIES.length()) {
+                var vmask = SPECIES.loadMask(a, i);
+                int tcount = vmask.trueCount();
+                int expectedTcount = 0;
+                for (int j = i; j < i + SPECIES.length(); j++) {
+                    expectedTcount += a[j] ? 1 : 0;
+                }
+                Assert.assertTrue(tcount == expectedTcount, "at index " + i + ", trueCount should be = " + expectedTcount + ", but is = " + tcount);
             }
-            Assert.assertTrue(tcount == expectedTcount, "at index " + i + ", trueCount should be = " + expectedTcount + ", but is = " + tcount);
         }
     }
 
@@ -430,17 +432,19 @@
     static void maskLastTrue$vectorteststype$SmokeTest(IntFunction<boolean[]> fa) {
         boolean[] a = fa.apply(SPECIES.length());
 
-        for (int i = 0; i < a.length; i += SPECIES.length()) {
-            var vmask = SPECIES.loadMask(a, i);
-            int ltrue = vmask.lastTrue();
-            int j = i + SPECIES.length() - 1;
-            for (; j >= i; j--) {
-                if (a[j]) break;
-            }
-            int expectedLtrue = j - i;
+        for (int ic = 0; ic < INVOC_COUNT * INVOC_COUNT; ic++) {
+            for (int i = 0; i < a.length; i += SPECIES.length()) {
+                var vmask = SPECIES.loadMask(a, i);
+                int ltrue = vmask.lastTrue();
+                int j = i + SPECIES.length() - 1;
+                for (; j >= i; j--) {
+                    if (a[j]) break;
+                }
+                int expectedLtrue = j - i;
 
-            Assert.assertTrue(ltrue == expectedLtrue, "at index " + i +
-                ", lastTrue should be = " + expectedLtrue + ", but is = " + ltrue);
+                Assert.assertTrue(ltrue == expectedLtrue, "at index " + i +
+                    ", lastTrue should be = " + expectedLtrue + ", but is = " + ltrue);
+            }
         }
     }
 
@@ -448,17 +452,19 @@
     static void maskFirstTrue$vectorteststype$SmokeTest(IntFunction<boolean[]> fa) {
         boolean[] a = fa.apply(SPECIES.length());
 
-        for (int i = 0; i < a.length; i += SPECIES.length()) {
-            var vmask = SPECIES.loadMask(a, i);
-            int ftrue = vmask.firstTrue();
-            int j = i;
-            for (; j < i + SPECIES.length() ; j++) {
-                if (a[j]) break;
-            }
-            int expectedFtrue = j - i;
+        for (int ic = 0; ic < INVOC_COUNT * INVOC_COUNT; ic++) {
+            for (int i = 0; i < a.length; i += SPECIES.length()) {
+                var vmask = SPECIES.loadMask(a, i);
+                int ftrue = vmask.firstTrue();
+                int j = i;
+                for (; j < i + SPECIES.length() ; j++) {
+                    if (a[j]) break;
+                }
+                int expectedFtrue = j - i;
 
-            Assert.assertTrue(ftrue == expectedFtrue, "at index " + i +
-                ", firstTrue should be = " + expectedFtrue + ", but is = " + ftrue);
+                Assert.assertTrue(ftrue == expectedFtrue, "at index " + i +
+                    ", firstTrue should be = " + expectedFtrue + ", but is = " + ftrue);
+            }
         }
     }
 
--- a/test/jdk/jdk/incubator/vector/templates/Unit-Miscellaneous.template
+++ b/test/jdk/jdk/incubator/vector/templates/Unit-Miscellaneous.template
@@ -422,11 +422,7 @@
     @Test(dataProvider = "maskProvider")
     static void maskTrueCount$vectorteststype$SmokeTest(IntFunction<boolean[]> fa) {
         boolean[] a = fa.apply(SPECIES.length());
-#if[Int]
-        int[] r = fr.apply(SPECIES.length());
-#else[Int]
-        int[] r = ifr.apply(SPECIES.length());
-#end[Int]
+        int[] r = new int[a.length];
 
         for (int ic = 0; ic < INVOC_COUNT * INVOC_COUNT; ic++) {
             for (int i = 0; i < a.length; i += SPECIES.length()) {
@@ -451,11 +447,7 @@
     @Test(dataProvider = "maskProvider")
     static void maskLastTrue$vectorteststype$SmokeTest(IntFunction<boolean[]> fa) {
         boolean[] a = fa.apply(SPECIES.length());
-#if[Int]
-        int[] r = fr.apply(SPECIES.length());
-#else[Int]
-        int[] r = ifr.apply(SPECIES.length());
-#end[Int]
+        int[] r = new int[a.length];
 
         for (int ic = 0; ic < INVOC_COUNT * INVOC_COUNT; ic++) {
             for (int i = 0; i < a.length; i += SPECIES.length()) {
@@ -480,11 +472,7 @@
     @Test(dataProvider = "maskProvider")
     static void maskFirstTrue$vectorteststype$SmokeTest(IntFunction<boolean[]> fa) {
         boolean[] a = fa.apply(SPECIES.length());
-#if[Int]
-        int[] r = fr.apply(SPECIES.length());
-#else[Int]
-        int[] r = ifr.apply(SPECIES.length());
-#end[Int]
+        int[] r = new int[a.length];
 
         for (int ic = 0; ic < INVOC_COUNT * INVOC_COUNT; ic++) {
             for (int i = 0; i < a.length; i += SPECIES.length()) {
--- a/test/jdk/jdk/incubator/vector/templates/Unit-Miscellaneous.template
+++ b/test/jdk/jdk/incubator/vector/templates/Unit-Miscellaneous.template
@@ -411,61 +411,89 @@
         }
     }
 
+    static int maskTrueCount(boolean[] a, int idx) {
+        int trueCount = 0;
+        for (int i = idx; i < idx + SPECIES.length(); i++) {
+            trueCount += a[i] ? 1 : 0;
+        }
+        return trueCount;
+    }
+
     @Test(dataProvider = "maskProvider")
     static void maskTrueCount$vectorteststype$SmokeTest(IntFunction<boolean[]> fa) {
         boolean[] a = fa.apply(SPECIES.length());
+#if[Int]
+        int[] r = fr.apply(SPECIES.length());
+#else[Int]
+        int[] r = ifr.apply(SPECIES.length());
+#end[Int]
 
         for (int ic = 0; ic < INVOC_COUNT * INVOC_COUNT; ic++) {
             for (int i = 0; i < a.length; i += SPECIES.length()) {
                 var vmask = SPECIES.loadMask(a, i);
-                int tcount = vmask.trueCount();
-                int expectedTcount = 0;
-                for (int j = i; j < i + SPECIES.length(); j++) {
-                    expectedTcount += a[j] ? 1 : 0;
-                }
-                Assert.assertTrue(tcount == expectedTcount, "at index " + i + ", trueCount should be = " + expectedTcount + ", but is = " + tcount);
+                r[i] = vmask.trueCount();
             }
         }
+
+        assertMaskReductionArraysEquals(r, a, $vectorteststype$::maskTrueCount);
+    }
+
+    static int maskLastTrue(boolean[] a, int idx) {
+        int i = idx + SPECIES.length() - 1;
+        for (; i >= idx; i--) {
+            if (a[i]) {
+                break;
+            }
+        }
+        return i - idx;
     }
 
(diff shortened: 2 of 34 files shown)


DamonFool:
How about the running time of these tests after the change?
Thanks.

DamonFool:
The improved tests also find more bugs of vector mask operations on x86.
Would you like to fix them together?

PaulSandoz (commented):
IIRC we wrote them as smoke tests because they were not intrinsic. We need to think more carefully about converting them from smoke tests.

Ideally we should convert them to kernel tests, but that is more work. Instead we can copy the generated pattern and do the following:

- move the assertion outside of the loops (it will generate garbage with string concatenation)
- assert over arrays, thereby also moving the actual scalar computation result outside the loops. The simplest approach is to create an `int[] array` of the same length as the input and write the reduced result at index `i`. Thus it's sparse.

That should result in an inner loop body that is very focused on exercising the intrinsic method. It will also likely reduce the test execution times.

dean-long (approved):
Seems reasonable.

XiaohongGong:
> How about the running time of these tests after the change?
> Thanks.

I tested the running time, and seems it almost have nothing influence to the whole tests running time. Thanks!

XiaohongGong:
> The improved tests also find more bugs of vector mask operations on x86.
> Would you like to fix them together?

Hi @DamonFool , thanks for looking at this PR! I also met the tests failures on some x86 machine. Seems something is wrong with the backend implementation. However, unfortunately I'm not very familiar with x86 instructions. So maybe @jatin-bhateja could help to take a look please? Thanks so much!

XiaohongGong:
> IIRC we wrote them as smoke tests because they were not intrinsic. We need to think more carefully about converting them from smoke tests.
> 
Thanks for looking at this PR @PaulSandoz ! Yes, I think this need more work and more carefully to move them from the smoke tests. Maybe we can revisit them in future?

> Ideally we should convert them to kernel tests, but that is more work. Instead we can copy the generated pattern and do the following:
> 
> * move the assertion outside of the loops (it will generate garbage with string concatenation)
> * assert over arrays, thereby also moving the actual scalar computation result outside the loops. The simplest approach is to create an `int[] array` of the same length as the input and write the reduced result at index `i`. Thus it's sparse.
> 
> That should result in an inner loop body that is very focused on exercising the intrinsic method. It will also likely reduce the test execution times.

Agree, I will move the assertion outside of the loop first. Thanks so much!

DamonFool:
> > The improved tests also find more bugs of vector mask operations on x86.
> > Would you like to fix them together?
> 
> Hi @DamonFool , thanks for looking at this PR! I also met the tests failures on some x86 machine. Seems something is wrong with the backend implementation. However, unfortunately I'm not very familiar with x86 instructions. So maybe @jatin-bhateja could help to take a look please? Thanks so much!

After some investigation, I believe the failure is caused by the incorrect code-gen for VectorStoreMask on x86.

Something like this would fix the failures:
```
diff --git a/src/hotspot/cpu/x86/x86.ad b/src/hotspot/cpu/x86/x86.ad
index ed34ab7..a837549 100644
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -7545,29 +7545,32 @@ instruct vstoreMask2B_evex(vec dst, vec src, immI_2 size) %{
   ins_pipe( pipe_slow );
 %}

-instruct storeMask4B(vec dst, vec src, immI_4 size) %{
+instruct storeMask4B(vec dst, vec src, immI_4 size, vec tmp) %{
   predicate (vector_length(n) <= 4 && UseAVX <= 2);
   match(Set dst (VectorStoreMask src size));
   format %{ "vector_store_mask $dst,$src\t!" %}
+  effect(TEMP dst, TEMP tmp);
   ins_encode %{
     assert(UseSSE >= 3, "required");
+    __ pxor($tmp$$XMMRegister, $tmp$$XMMRegister);
     __ pabsd($dst$$XMMRegister, $src$$XMMRegister);
-    __ packssdw($dst$$XMMRegister, $dst$$XMMRegister);
-    __ packsswb($dst$$XMMRegister, $dst$$XMMRegister);
+    __ packssdw($dst$$XMMRegister, $tmp$$XMMRegister);
+    __ packsswb($dst$$XMMRegister, $tmp$$XMMRegister);
   %}
   ins_pipe( pipe_slow );
 %}

-instruct vstoreMask4B(vec dst, vec src, immI_4 size) %{
+instruct vstoreMask4B(vec dst, vec src, immI_4 size, vec tmp) %{
   predicate(vector_length(n) == 8 && UseAVX <= 2);
   match(Set dst (VectorStoreMask src size));
   format %{ "vector_store_mask $dst,$src\t!" %}
-  effect(TEMP dst);
+  effect(TEMP dst, TEMP tmp);
   ins_encode %{
     int vlen_enc = Assembler::AVX_128bit;
+    __ pxor($tmp$$XMMRegister, $tmp$$XMMRegister);
     __ vextracti128($dst$$XMMRegister, $src$$XMMRegister, 0x1);
     __ vpackssdw($dst$$XMMRegister, $src$$XMMRegister, $dst$$XMMRegister, vlen_enc);
-    __ vpacksswb($dst$$XMMRegister, $dst$$XMMRegister, $dst$$XMMRegister, vlen_enc);
+    __ vpacksswb($dst$$XMMRegister, $dst$$XMMRegister, $tmp$$XMMRegister, vlen_enc);
     __ vpabsb($dst$$XMMRegister, $dst$$XMMRegister, vlen_enc);
   %}
   ins_pipe( pipe_slow );
@@ -7613,8 +7616,9 @@ instruct storeMask8B_avx(vec dst, vec src, immI_8 size, legVec vtmp) %{
     __ vpshufps($dst$$XMMRegister, $src$$XMMRegister, $src$$XMMRegister, 0x88, Assembler::AVX_256bit);
     __ vextracti128($vtmp$$XMMRegister, $dst$$XMMRegister, 0x1);
     __ vblendps($dst$$XMMRegister, $dst$$XMMRegister, $vtmp$$XMMRegister, 0xC, vlen_enc);
-    __ vpackssdw($dst$$XMMRegister, $dst$$XMMRegister, $dst$$XMMRegister, vlen_enc);
-    __ vpacksswb($dst$$XMMRegister, $dst$$XMMRegister, $dst$$XMMRegister, vlen_enc);
+    __ pxor($vtmp$$XMMRegister, $vtmp$$XMMRegister);
+    __ vpackssdw($dst$$XMMRegister, $dst$$XMMRegister, $vtmp$$XMMRegister, vlen_enc);
+    __ vpacksswb($dst$$XMMRegister, $dst$$XMMRegister, $vtmp$$XMMRegister, vlen_enc);
     __ vpabsb($dst$$XMMRegister, $dst$$XMMRegister, vlen_enc);
   %}
   ins_pipe( pipe_slow );
```

But I'm afraid there are more rules which need to be fixed.
So I've filed JDK-8269679.

Please feel free to go ahead.
Thanks.

dean-long (changes requested):
I am seeing testing failures.

sviswa7:
@XiaohongGong The following should fix this for x86:

diff --git a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
index 147bcb8..b548877 100644
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -3870,6 +3870,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vpsubb(xtmp, xtmp, mask, vec_enc);
   evpmovb2m(ktmp, xtmp, vec_enc);
   kmovql(tmp, ktmp);
+  if (masklen < 64) {
+    andq(tmp, (((jlong)1 << masklen) - 1));
+  }
   switch(opc) {
     case Op_VectorMaskTrueCount:
       popcntq(dst, tmp);
@@ -3894,6 +3897,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vpxor(xtmp, xtmp, xtmp, vec_enc);
   vpsubb(xtmp, xtmp, mask, vec_enc);
   vpmovmskb(tmp, xtmp, vec_enc);
+  if (masklen < 64) {
+    andq(tmp, (((jlong)1 << masklen) - 1));
+  }
   switch(opc) {
     case Op_VectorMaskTrueCount:
       popcntq(dst, tmp);

Please include this in your PR.

DamonFool:
> @XiaohongGong The following should fix this for x86:
> 
> ```
> diff --git a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
> index 147bcb8..b548877 100644
> --- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
> +++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
> @@ -3870,6 +3870,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
>    vpsubb(xtmp, xtmp, mask, vec_enc);
>    evpmovb2m(ktmp, xtmp, vec_enc);
>    kmovql(tmp, ktmp);
> +  if (masklen < 64) {
> +    andq(tmp, (((jlong)1 << masklen) - 1));
> +  }
>    switch(opc) {
>      case Op_VectorMaskTrueCount:
>        popcntq(dst, tmp);
> @@ -3894,6 +3897,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
>    vpxor(xtmp, xtmp, xtmp, vec_enc);
>    vpsubb(xtmp, xtmp, mask, vec_enc);
>    vpmovmskb(tmp, xtmp, vec_enc);
> +  if (masklen < 64) {
> +    andq(tmp, (((jlong)1 << masklen) - 1));
> +  }
>    switch(opc) {
>      case Op_VectorMaskTrueCount:
>        popcntq(dst, tmp);
> ```
> 
> Please include this in your PR.

The fix seems to be tricky.
So do you think we need to change the code-gen like this https://github.com/openjdk/jdk17/pull/168#issuecomment-871412973 ?  @sviswa7

sviswa7:
@DamonFool This is the cleanest solution I could think of.
The problem comes only with newly added intrinsic where firstTrue, lastTrue, trueCount become consumers to VectorStoreMask. All the other consumers of VectorStoreMask consume the right number of bits. 
The fix is to make the new intrinsic also consume the right number of bits by doing the 'and' masking.

DamonFool:
> @DamonFool This is the cleanest solution I could think of.
> The problem comes only with newly added intrinsic where firstTrue, lastTrue, trueCount become consumers to VectorStoreMask. All the other consumers of VectorStoreMask consume the right number of bits.
> The fix is to make the new intrinsic also consume the right number of bits by doing the 'and' masking.

But we actually do not `and` masking with AVX3.
So I'm afraid the root cause is the incorrect code-gen for VectorStoreMask.

sviswa7:
@DamonFool This small patch fixes the problem at hand and at the minimum should go in with JDK 17. You have filed the JDK-8269679 for a more robust fix.  Also the masking support is being overhauled as part of JDK 18. Your thoughts?

DamonFool:
> @DamonFool This small patch fixes the problem at hand and at the minimum should go in with JDK 17. You have filed the JDK-8269679 for a more robust fix. Also the masking support is being overhauled as part of JDK 18. Your thoughts?

Actually, instruct like `storeMask1B` and `storeMask2B` don't need `and masking` for both AVX2 and AVX3.
But to lower the risk of JDK17, I agree with you.
Thanks.

sviswa7:
Thanks @DamonFool.

XiaohongGong:
> @XiaohongGong The following should fix this for x86:
> 
> ```
> diff --git a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
> index 147bcb8..b548877 100644
> --- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
> +++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
> @@ -3870,6 +3870,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
>    vpsubb(xtmp, xtmp, mask, vec_enc);
>    evpmovb2m(ktmp, xtmp, vec_enc);
>    kmovql(tmp, ktmp);
> +  if (masklen < 64) {
> +    andq(tmp, (((jlong)1 << masklen) - 1));
> +  }
>    switch(opc) {
>      case Op_VectorMaskTrueCount:
>        popcntq(dst, tmp);
> @@ -3894,6 +3897,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
>    vpxor(xtmp, xtmp, xtmp, vec_enc);
>    vpsubb(xtmp, xtmp, mask, vec_enc);
>    vpmovmskb(tmp, xtmp, vec_enc);
> +  if (masklen < 64) {
> +    andq(tmp, (((jlong)1 << masklen) - 1));
> +  }
>    switch(opc) {
>      case Op_VectorMaskTrueCount:
>        popcntq(dst, tmp);
> ```
> 
> Please include this in your PR.

Thanks for the fix @sviswa7 ! I will include it in the latest commit. Thanks!

XiaohongGong:
@PaulSandoz , I'v updated the tests part. Could you please check again whether it make sense to you? Thanks so much!

XiaohongGong:
Hi @sviswa7 @DamonFool @dean-long , the x86 issue is fixed with @sviswa7 's patch.  I'v tested the Vector API tests internally and they all pass. Could you please take a look at it again? Thanks so much!

DamonFool (approved):
Changes in HotSpot LGTM.
Thanks for fixing it.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -3887,6 +3890,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vpxor(xtmp, xtmp, xtmp, vec_enc);
   vpsubb(xtmp, xtmp, mask, vec_enc);
   vpmovmskb(tmp, xtmp, vec_enc);
+  if (masklen < 64) {
+    andq(tmp, (((jlong)1 << masklen) - 1));
+  }
```
jatin-bhateja, line 3895: Hi @sviswa7 , @XiaohongGong , I think problem will occur only on non-AVX512VLBW platforms. So this workaround fix can be removed from this place.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -3863,6 +3863,9 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vpsubb(xtmp, xtmp, mask, vec_enc);
   evpmovb2m(ktmp, xtmp, vec_enc);
   kmovql(tmp, ktmp);
+  if (masklen < 64) {
+    andq(tmp, (((jlong)1 << masklen) - 1));
+  }
```
jatin-bhateja, line 3868: Hi @sviswa7 , @XiaohongGong , @DamonFool , I think problem will occur only on non-AVX512VLBW platforms. So this workaround fix can be removed from this place.

test/jdk/jdk/incubator/vector/templates/Unit-Miscellaneous.template
```diff
@@ -411,55 +411,89 @@
...
+        for (int i = idx; i < idx + SPECIES.length(); i++) {
+            trueCount += a[i] ? 1 : 0;
+        }
+        return trueCount;
+    }
+
     @Test(dataProvider = "maskProvider")
     static void maskTrueCount$vectorteststype$SmokeTest(IntFunction<boolean[]> fa) {
         boolean[] a = fa.apply(SPECIES.length());
+#if[Int]
```
PaulSandoz, line 425: We already know the length of `r`, it's `a.length`, so we can just do:
```java 
        int[] r = new int[a.length];
```
Thereby reducing the complexity. I realize you are copying the pattern of `lfr` which i think is also unnecessary. Let's clean this up later.

PaulSandoz (approved):
Very nice update to the test.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
sviswa7: @jatin-bhateja Thanks for this input.
@XiaohongGong Please include the following patch for Jatin's comment:
```
diff --git a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
index 87016e2..e21b065 100644
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -3863,9 +3863,6 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vpsubb(xtmp, xtmp, mask, vec_enc);
   evpmovb2m(ktmp, xtmp, vec_enc);
   kmovql(tmp, ktmp);
-  if (masklen < 64) {
-    andq(tmp, (((jlong)1 << masklen) - 1));
-  }
   switch(opc) {
     case Op_VectorMaskTrueCount:
       popcntq(dst, tmp);
```

test/jdk/jdk/incubator/vector/templates/Unit-Miscellaneous.template
XiaohongGong: Resolved. Thanks for the suggestion!

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
XiaohongGong: Removed!

XiaohongGong:
Hi @sviswa7 @PaulSandoz , all your comments have been addressed! Thanks so much for the review!

jatin-bhateja (approved):

sviswa7 (approved):

dean-long:
I'm seeing some tests timing out.  @PaulSandoz how did your testing go?

PaulSandoz:
> I'm seeing some tests timing out. @PaulSandoz how did your testing go?

For me there were no timeouts Can you provide more details?

PaulSandoz:
> > I'm seeing some tests timing out. @PaulSandoz how did your testing go?
> 
> For me there were no timeouts Can you provide more details?

Vector tests for run as part of hs-tier1, hs-tier2, and hs-tier3 are fine. You might be observing time outs in unrelated tests.

dean-long:
The timeouts were in 4 different jdk/incubator/vector tests, but all were on the same machine linux-aarch64 machine, and all were fastdebug builds, tier3, with ZGC.

XiaohongGong:
> The timeouts were in 4 different jdk/incubator/vector tests, but all were on the same machine linux-aarch64 machine, and all were fastdebug builds, tier3, with ZGC.

Do these tests also time out without this patch? Could you please provider more details to the tests (the timing out log or the test itself...). I'm not sure, but I guess the timing out is not related to this patch?

XiaohongGong:
> > The timeouts were in 4 different jdk/incubator/vector tests, but all were on the same machine linux-aarch64 machine, and all were fastdebug builds, tier3, with ZGC.
> 
> Do these tests also time out without this patch? Could you please provider more details to the tests (the timing out log or the test itself...). I'm not sure, but I guess the timing out is not related to this patch?

I ran the Vector API tests with ZGC on all kinds of aarch64 machines that we have, and all the tests pass without the timing out.

dean-long:
It's probably just a glitch on one of our test machines.  I checked the results again and the tests ran fine with ZGC on other linux-aarch64 test machines.

XiaohongGong:
> It's probably just a glitch on one of our test machines. I checked the results again and the tests ran fine with ZGC on other linux-aarch64 test machines.

OK, that's very fine! Thanks so much!

PR #169, opened 2021-06-29
8269285: Crash/miscompile in CallGenerator::for_method_handle_inline after JDK-8191998

shipilev:
See the bug report for more details.

I believe the [JDK-8191998](https://bugs.openjdk.java.net/browse/JDK-8191998) change introduced a slight regression, where the speculative type join may empty the type. It would then crash on assert in `fastdebug` builds, or miscompile the null-check to `true` in `release` bits. New test captures both failure modes.

This is not a recent regression, but a regression nevertheless, so I would like to have that fix in JDK 17. Please review carefully, or speak up if you want to move it to JDK 18+ and then backport later.

Additional testing:
 - [x] New test fails without the patch, passes with it
 - [x] Linux x86_64 `fastdebug` `tier1`

---------
### Progress

### Issue
 * [JDK-8269285](https://bugs.openjdk.java.net/browse/JDK-8269285): Crash/miscompile in CallGenerator::for_method_handle_inline after JDK-8191998

### Reviewers
 * [Igor Veresov](https://openjdk.java.net/census#iveresov) (@veresov - **Reviewer**) ⚠️ Review applies to 86b6f2cfa19356ffb505296531749171833d10aa
 * [Vladimir Ivanov](https://openjdk.java.net/census#vlivanov) (@iwanowww - **Reviewer**) ⚠️ Review applies to 02e01bbd93717110f39ef4c57b6ea7deeeb82171

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/169/head:pull/169` \
`$ git checkout pull/169`

`$ git checkout pull/169` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/169/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 169`

`$ git pr show -t 169`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/169.diff">https://git.openjdk.java.net/jdk17/pull/169.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1151,8 +1151,10 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
           const Type*       sig_type = TypeOopPtr::make_from_klass(signature->accessing_klass());
           if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-            const Type* recv_type = arg_type->join_speculative(sig_type); // keep speculative part
-            Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, recv_type));
+            // Keep the speculative part, but disallow it to empty the type
+            const Type* recv_type = arg_type->join_speculative(sig_type);
+            const Type* cast_type = recv_type->empty() ? sig_type : recv_type;
+            Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, cast_type));
             kit.set_argument(0, cast_obj);
           }
         }
@@ -1164,8 +1166,10 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
             const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass());
             if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-              const Type* narrowed_arg_type = arg_type->join_speculative(sig_type); // keep speculative part
-              Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, narrowed_arg_type));
+              // Keep the speculative part, but disallow it to empty the type
+              const Type* narrowed_arg_type = arg_type->join_speculative(sig_type);
+              const Type* cast_type = narrowed_arg_type->empty() ? sig_type : narrowed_arg_type;
+              Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, cast_type));
               kit.set_argument(receiver_skip + j, cast_obj);
             }
           }
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1148,13 +1148,14 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
         // Cast receiver to its type.
         if (!target->is_static()) {
           Node* arg = kit.argument(0);
+          ciKlass* sig_klass = signature->accessing_klass();
           const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
-          const Type*       sig_type = TypeOopPtr::make_from_klass(signature->accessing_klass());
+          const Type*       sig_type = TypeOopPtr::make_from_klass(sig_klass);
           if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-            // Keep the speculative part, but disallow it to empty the type
-            const Type* recv_type = arg_type->join_speculative(sig_type);
-            const Type* cast_type = recv_type->empty() ? sig_type : recv_type;
-            Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, cast_type));
+            // Keep the speculative part, but do not join with interfaces
+            const Type* recv_type = sig_klass->is_interface() ?
+                    sig_type : arg_type->join_speculative(sig_type);
+            Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, recv_type));
             kit.set_argument(0, cast_obj);
           }
         }
@@ -1163,13 +1164,14 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           ciType* t = signature->type_at(i);
           if (t->is_klass()) {
             Node* arg = kit.argument(receiver_skip + j);
+            ciKlass* sig_klass = t->as_klass();
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
-            const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass());
+            const Type*       sig_type = TypeOopPtr::make_from_klass(sig_klass);
             if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-              // Keep the speculative part, but disallow it to empty the type
-              const Type* narrowed_arg_type = arg_type->join_speculative(sig_type);
-              const Type* cast_type = narrowed_arg_type->empty() ? sig_type : narrowed_arg_type;
-              Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, cast_type));
+              // Keep the speculative part, but do not join with interfaces
+              const Type* narrowed_arg_type = sig_klass->is_interface() ?
+                      sig_type : arg_type->join_speculative(sig_type);
+              Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, narrowed_arg_type));
               kit.set_argument(receiver_skip + j, cast_obj);
             }
           }
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1151,7 +1151,7 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
           const Type*       sig_type = TypeOopPtr::make_from_klass(signature->accessing_klass());
           if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-            const Type* recv_type = arg_type->filter_speculative(sig_type); // keep speculative parts
+            const Type* recv_type = arg_type->filter_speculative(sig_type); // keep speculative part
             Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, recv_type));
             kit.set_argument(0, cast_obj);
           }
@@ -1164,7 +1164,7 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
             const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass());
             if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-              const Type* narrowed_arg_type = arg_type->filter_speculative(sig_type); // keep speculative parts
+              const Type* narrowed_arg_type = arg_type->filter_speculative(sig_type); // keep speculative part
               Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, narrowed_arg_type));
               kit.set_argument(receiver_skip + j, cast_obj);
             }
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1148,13 +1148,10 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
         // Cast receiver to its type.
         if (!target->is_static()) {
           Node* arg = kit.argument(0);
-          ciKlass* sig_klass = signature->accessing_klass();
           const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
-          const Type*       sig_type = TypeOopPtr::make_from_klass(sig_klass);
+          const Type*       sig_type = TypeOopPtr::make_from_klass(signature->accessing_klass());
           if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-            // Keep the speculative part, but do not join with interfaces
-            const Type* recv_type = sig_klass->is_interface() ?
-                    sig_type : arg_type->join_speculative(sig_type);
+            const Type* recv_type = arg_type->filter_speculative(sig_type); // keep speculative parts
             Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, recv_type));
             kit.set_argument(0, cast_obj);
           }
@@ -1164,13 +1161,10 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           ciType* t = signature->type_at(i);
           if (t->is_klass()) {
             Node* arg = kit.argument(receiver_skip + j);
-            ciKlass* sig_klass = t->as_klass();
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
-            const Type*       sig_type = TypeOopPtr::make_from_klass(sig_klass);
+            const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass());
             if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-              // Keep the speculative part, but do not join with interfaces
-              const Type* narrowed_arg_type = sig_klass->is_interface() ?
-                      sig_type : arg_type->join_speculative(sig_type);
+              const Type* narrowed_arg_type = arg_type->filter_speculative(sig_type); // keep speculative parts
               Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, narrowed_arg_type));
               kit.set_argument(receiver_skip + j, cast_obj);
             }
--- a/test/hotspot/jtreg/compiler/types/TestMethodHandleSpeculation.java
+++ b/test/hotspot/jtreg/compiler/types/TestMethodHandleSpeculation.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2021, Red Hat, Inc.. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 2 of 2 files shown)


vnkozlov:
@shipilev can it be related to  [8268884](https://bugs.openjdk.java.net/browse/JDK-8268884)?

shipilev:
> @shipilev can it be related to [8268884](https://bugs.openjdk.java.net/browse/JDK-8268884)?

It does not look like it to me, but I am not the expert in this code. I thought this one is similar to the code [in other place](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/opto/memnode.cpp#L1930-L1935). Anyhow, applying the patch from JDK-8268884 PR does not fix the new regression test from this PR.

shipilev:
@vnkozlov More precisely, here are the types that `join_speculative` encounters on the failing path:

```
STDOUT:
arg_type: byte[int:>=0]:exact * (speculative=byte[int:>=0]:NotNull:exact *)
sig_type: java/io/Serializable *
narrowed_arg_type: java/lang/Object:AnyNull *,iid=top
```

veresov:
I'm not an expect on this but perhaps we shouldn't be doing a join but instead transfer the speculative part of the argument to the type of the signature?

shipilev:
> I'm not an expect on this but perhaps we shouldn't be doing a join and instead transfer the speculative part of the argument to the speculative part of type of the signature?

I thought `join_speculative` was the attempt at that; but that fails when argument and signature are seriously different (interface and boxed class (?) here). It seems some other uses of `join_speculative` appreciate that corner case and handle it. I would be happy to try other approaches, but cannot find an easy (as in clean and obvious) way to do this.

veresov:
Well, `join_speculative()` also does a join between the normal types. And evidently the type system has no idea how `byte[]` relates to `Serializable`, so the answer is the lowest type possible, which is `Object`. I hope Roland cleans this up in his type system rework down the road.

To do the surgery on the speculative part, it seems like the only way to do that is to add a `Type* cast_with_speculative(Type*)` to the `Type` interface. Otherwise I don't see how we can clone all the little protected fields in `TypeOopPtr`. Or may be some idiom exists that already does that... But I don't know. @vnkozlov, what do you think?

shipilev:
Right. Is it odd that `join_speculative` produces `AnyNull` here? Anyhow, given the miscompilation is already observed in the wild, maybe the current patch is a good protective measure for the cases where Type join fails?

veresov:
`AnyNull` seems odd, but that's what makes you fix work, making `empty()` return true.

veresov:
Come to think of it, this pathology arises because one of these guys is an interface. The type system doesn't quite know how to deal with that. Perhaps we should guard for it and not attempt the join if the argument type is an interface?

vnkozlov (approved):
I think for JDK 17 it is good fix. It revert result type to one before 8191998 changes in case `join` failed. In this sense it is not worser then before 8191998. We can look more on this later. May be when Vladimir Ivanov back.

vnkozlov:
About joining with Interface type. Returning `Object:AnyNull` is normal answer from C2 for Interfaces. We can't handle them currently and it is not the goal of this patch.

veresov:
Ok, if `AnyNull` is what is reliably returned for interfaces, then `empty()` is essentially equivalent to `is_interface()` in this case. Thanks for explanation.

veresov (approved):

iwanowww:
I'm not fond of the proposed fix. Though it works in this particular case, `Type::empty()` may give a wrong signal in some pathological/corner cases.

As an example,`join`/`join_speculative` may produce `TOP` when call site happens to be in effectively dead (or dying) part of the graph and paradoxical IR shapes can arise. In such case, there's no reason to continue inlining through the call site, but give up (return `NULL`)  immediately.

In this particular case (`sig_type` represents an interface), it goes off the rails when `arg_type->higher_equal(sig_type)` returns `false` though arrays always implement `Serializable`. The cast node is redundant here and can be simply omitted. (Recently, JDK-8268371 addressed a similar problem uncovered by CHA-related changes.)

I'm in favor of explicitly filtering out interface types coming from signatures and simply omitting casts when interface is encountered.

shipilev:
> I'm in favor of explicitly filtering out interface types coming from signatures and simply omitting casts when interface is encountered.

All right, would this be acceptable then?

```
diff --git a/src/hotspot/share/opto/callGenerator.cpp b/src/hotspot/share/opto/callGenerator.cpp
index ab7f7897797..5da828ebae9 100644
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1148,9 +1148,11 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
         // Cast receiver to its type.
         if (!target->is_static()) {
           Node* arg = kit.argument(0);
+          ciKlass* sig_klass = signature->accessing_klass();
           const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
-          const Type*       sig_type = TypeOopPtr::make_from_klass(signature->accessing_klass());
-          if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
+          const Type*       sig_type = TypeOopPtr::make_from_klass(sig_klass);
+          if (arg_type != NULL && !arg_type->higher_equal(sig_type) &&
+              !sig_klass->is_interface()) { // avoid joins with interfaces
             const Type* recv_type = arg_type->join_speculative(sig_type); // keep speculative part
             Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, recv_type));
             kit.set_argument(0, cast_obj);
@@ -1161,9 +1163,10 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           ciType* t = signature->type_at(i);
           if (t->is_klass()) {
             Node* arg = kit.argument(receiver_skip + j);
+            ciKlass* sig_klass = t->as_klass();
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
-            const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass());
-            if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
+            const Type*       sig_type = TypeOopPtr::make_from_klass(sig_klass);
+            if (arg_type != NULL && !arg_type->higher_equal(sig_type) && !sig_klass->is_interface()) {
               const Type* narrowed_arg_type = arg_type->join_speculative(sig_type); // keep speculative part
               Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, narrowed_arg_type));
               kit.set_argument(receiver_skip + j, cast_obj);
```

shipilev:
> I'm in favor of explicitly filtering out interface types coming from signatures and simply omitting casts when interface is encountered.

All right, would this be acceptable then?

```diff
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1148,9 +1148,10 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
         // Cast receiver to its type.
         if (!target->is_static()) {
           Node* arg = kit.argument(0);
+          ciKlass* sig_klass = signature->accessing_klass();
           const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
-          const Type*       sig_type = TypeOopPtr::make_from_klass(signature->accessing_klass());
-          if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
+          const Type*       sig_type = TypeOopPtr::make_from_klass(sig_klass);
+          if (arg_type != NULL && !arg_type->higher_equal(sig_type) && !sig_klass->is_interface()) {
             const Type* recv_type = arg_type->join_speculative(sig_type); // keep speculative part
             Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, recv_type));
             kit.set_argument(0, cast_obj);
@@ -1161,9 +1162,10 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           ciType* t = signature->type_at(i);
           if (t->is_klass()) {
             Node* arg = kit.argument(receiver_skip + j);
+            ciKlass* sig_klass = t->as_klass();
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
-            const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass());
-            if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
+            const Type*       sig_type = TypeOopPtr::make_from_klass(sig_klass);
+            if (arg_type != NULL && !arg_type->higher_equal(sig_type) && !sig_klass->is_interface()) {
               const Type* narrowed_arg_type = arg_type->join_speculative(sig_type); // keep speculative part
               Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, narrowed_arg_type));
               kit.set_argument(receiver_skip + j, cast_obj);
```

shipilev:
What I am concerned about with filtering interfaces in signatures, is that we are dealing with `MethodHandle` handling block. Would  such filtering break when `arg_type` and `sig_type` are both interfaces?

shipilev:
Wait, actually I should just filter before calling `join_speculative`, which makes my concern moot. See new commit It passes the regression test, I'll run `tier1`, `tier2` now...

iwanowww:
> Would such filtering break when arg_type and sig_type are both interfaces?

What particular scenario do you have in mind? The check is performend against `sig_klass` only, so it covers the case when both `arg_type` and `sig_type` represent interfaces. Interface types aren't trusted in bytecode, so the worst case scenario is there'll be a repeated subtype check at runtime needed later in callee. 

MethodHandle invocation performs an adhoc type check between call site signature and MethodHandle type (`MH.type() == MT_call_site`) and the relevant logic in `CallGenerator::for_method_handle_inline()` reifies its effects on argument types in IR.

src/hotspot/share/opto/callGenerator.cpp
```diff
@@ -1161,10 +1164,13 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           ciType* t = signature->type_at(i);
           if (t->is_klass()) {
```
iwanowww, line 1165: What I suggested is to extend the guard with `!t->as_klass()->is_interface()` check.

shipilev: I am not convinced it is safe. At least it does not look safe for JDK 17. The "must" in the comment block above makes me wary about skipping the casting for interfaces...

```
        // In lambda forms we erase signature types to avoid resolving issues
        // involving class loaders.  When we optimize a method handle invoke
        // to a direct call we must cast the receiver and arguments to its
        // actual types.
```

shipilev:
OK, so there are basically three ways to fix it so far. In the draft form, they are:

1) Handle `empty()` `join` result:

```diff
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1164,8 +1166,10 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
             const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass());
             if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-              const Type* narrowed_arg_type = arg_type->join_speculative(sig_type); // keep speculative part
-              Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, narrowed_arg_type));
+              // Keep the speculative part, but disallow it to empty the type
+              const Type* narrowed_arg_type = arg_type->join_speculative(sig_type);
+              const Type* cast_type = narrowed_arg_type->empty() ? sig_type : narrowed_arg_type;
+              Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, cast_type));
               kit.set_argument(receiver_skip + j, cast_obj);
             }
           }
```

2) Filter `is_interface` before performing the `join`, but still attempt to cast.

```diff
diff --git a/src/hotspot/share/opto/callGenerator.cpp b/src/hotspot/share/opto/callGenerator.cpp
index ab7f7897797..eaf19d37b72 100644
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1161,10 +1164,13 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           ciType* t = signature->type_at(i);
           if (t->is_klass()) {
             Node* arg = kit.argument(receiver_skip + j);
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
            const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass()));
             if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-              const Type* narrowed_arg_type = arg_type->join_speculative(sig_type); // keep speculative part
+              // Keep the speculative part, but do not join with interfaces
+              const Type* narrowed_arg_type = t->as_klass()->is_interface() ?
+                      sig_type : arg_type->join_speculative(sig_type);
               Node* cast_obj = gvn.transform(new CheckCastPPNode(kit.control(), arg, narrowed_arg_type));
               kit.set_argument(receiver_skip + j, cast_obj);
             }
```

3) Filter `is_interface` arguments completely, i.e. do not even attempt the cast.

```diff
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -1159,7 +1159,7 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
         // Cast reference arguments to its type.
         for (int i = 0, j = 0; i < signature->count(); i++) {
           ciType* t = signature->type_at(i);
-          if (t->is_klass()) {
+          if (t->is_klass() && !t->as_klass()->is_interface()) {
             Node* arg = kit.argument(receiver_skip + j);
             const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
             const Type*       sig_type = TypeOopPtr::make_from_klass(t->as_klass());
```

All three fix the regression test. All three pass `tier1`, `tier2`.

I think (2) is the least risky for JDK 17. Thoughts?

src/hotspot/share/opto/callGenerator.cpp
iwanowww: Not sure what safety concerns you have in mind w.r.t interfaces. On bytecode level interface treatment is lax (e.g., verifier ignores interface type info; see JVMS-4.10.1.2 for details).

shipilev: I am mostly concerned about pushing a "less than obvious and wider than definitely required" fix to JDK 17. Filtering out by "empty join type" is quite probably too wide. Filtering out the interface arguments is much less obvious. Yes, I get that interface arguments in signatures might be filtered wholesale, but I would rather wait for more testing to see if that actually works.

iwanowww: Frankly speaking, I wouldn't expect too much from testing considering the problem went unnoticed for 2,5 years.

But, as an additional data point, I'd like to reiterate that the fix for JDK-8268371 applied a similar strategy in a much more common scenario - inlining of interface methods - and so far there were no problems spotted.

iwanowww:
Also, another way to fix the immediate bug is to switch from `join_speculative` to `filter_speculative` which, among other things, tries to work around limitations imposed by interface types:
```
> p arg_type->dump()
byte[int:>=0]:exact * (speculative=byte[int:>=0]:NotNull:exact *)

> p sig_type->dump()
java/io/Serializable *

> p arg_type->join_speculative(sig_type)->dump()
java/lang/Object:AnyNull *,iid=top

> p arg_type->filter_speculative(sig_type)->dump()
java/io/Serializable *
```

veresov:
I like having `is_interface()` there because it makes it easier to find this place later, when the type system is fixed.

iwanowww:
After thinking more about it (and studying `TypeOopPtr::filter_helper`), switching to `filter_speculative` looks like the right way to fix the bug. 

Also, it won't require any additional changes when the type system is fixed.

shipilev:
Cool, `filter_speculative` works for regression test. See new commit. I'll run `tier{1,2}` for it.

iwanowww:
> Also, it won't require any additional changes when the type system is fixed.

I take it back: the sole need for joining `arg_type` and `sig_type` is to work around limitations imposed by `CheckCastPP`. And the sole purpose of `CheckCastPP` (compared to `CastPP`) is to work around limitations imposed by interface types. Once the type system is enhanced with proper support of interface types, I believe it will become possible to get rid of `CheckCastPP` in favor of `CastPP` and there won't be any need in explicit joining of types in `CallGenerator::for_method_handle_inline()` since `CastPPNode::Value()` will do "the right thing" and keep the speculative part along the way.

BTW I don't see any reason why `CheckCastPPNode::Value()` can't propagate speculative part of the input type right away, but that's something to consider as a followup enhancement.

src/hotspot/share/opto/callGenerator.cpp
```diff
@@ -1151,7 +1151,7 @@ CallGenerator* CallGenerator::for_method_handle_inline(JVMState* jvms, ciMethod*
           const TypeOopPtr* arg_type = arg->bottom_type()->isa_oopptr();
           const Type*       sig_type = TypeOopPtr::make_from_klass(signature->accessing_klass());
           if (arg_type != NULL && !arg_type->higher_equal(sig_type)) {
-            const Type* recv_type = arg_type->join_speculative(sig_type); // keep speculative part
+            const Type* recv_type = arg_type->filter_speculative(sig_type); // keep speculative parts
```
iwanowww, line 1154: No need to adjust the comment: `sig_type` doesn't have speculative part (by construction).

iwanowww (approved):
Looks good.

src/hotspot/share/opto/callGenerator.cpp
shipilev: Right, I restored the comment incompletely.

test/hotspot/jtreg/compiler/types/TestMethodHandleSpeculation.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+/*
+ * @test
+ * @bug 8269285
+ * @summary Crash/miscompile in CallGenerator::for_method_handle_inline after JDK-8191998
+ * @requires vm.compMode == "Xmixed" & vm.flavor == "server"
+ *
+ * @run main/othervm
+ *        -Xcomp -Xbatch
```
iwanowww, line 31: The test can be sped up significatly by compiling `compiler.types.TestMethodHandleSpeculation::main()` only (`-XX:CompileCommand=quiet -XX:CompileCommand=compileonly,compiler.types.TestMethodHandleSpeculation::main`).

shipilev: I tried it before with `-XX:CompileOnly=compiler.types.TestMethodHandleSpeculation::*`, and IIRC that did not work. I suspect you need to compile/inline the lambda body too. After all, this is `CallGenerator::for_method_handle_inline`? I can try again.

iwanowww: There's some difference in behavior between `-XX:CompileCommand=compileonly,` and `-XX:CompileOnly=`.
IIRC the former doesn't affect inlining while the latter does.

shipilev: Right, that works, hm. See update.

shipilev:
`tier1` and `tier2` are fine with `filter_speculative`. @vnkozlov, @veresov -- are you good with new version?

vnkozlov (approved):
Yes, looks good.

veresov:
Looks good to me.

shipilev:
Thank you, all.

/integrate

PR #172, opened 2021-06-29
8269580: assert(is_valid()) failed: invalid register (-1)

jatin-bhateja:
- Assertion fails during emit phase of rep_stos instruction pattern on targets not supporting AVX512BW feature.
- Pattern is selected under following predication logic.
 predicate(!((ClearArrayNode*)n)->is_large() &&
              (UseAVX <= 2 || !VM_Version::supports_avx512vlbw()));
- Encoding block of this pattern passes a  knoreg opmask register having a default encoding of -1,  this later causes an assertion failure while assembling instruction operating over this register. 
- Existing pattern rep_stos_evex should be able to handle case for AVX512 targets as instructions operating of opmask registers are anyways guarded by target feature checks.

---------
### Progress

### Issue
 * [JDK-8269580](https://bugs.openjdk.java.net/browse/JDK-8269580): assert(is_valid()) failed: invalid register (-1)

### Reviewers
 * [Dean Long](https://openjdk.java.net/census#dlong) (@dean-long - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/172/head:pull/172` \
`$ git checkout pull/172`

`$ git checkout pull/172` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/172/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 172`

`$ git pr show -t 172`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/172.diff">https://git.openjdk.java.net/jdk17/pull/172.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/x86_64.ad
+++ b/src/hotspot/cpu/x86/x86_64.ad
@@ -11024,7 +11024,7 @@ instruct rep_stos(rcx_RegL cnt, rdi_RegP base, regD tmp, rax_RegI zero,
                   Universe dummy, rFlagsReg cr)
 %{
   predicate(!((ClearArrayNode*)n)->is_large() &&
-              (UseAVX <= 2 || !VM_Version::supports_avx512vlbw()));
+              (UseAVX <= 2));
   match(Set dummy (ClearArray cnt base));
   effect(USE_KILL cnt, USE_KILL base, TEMP tmp, KILL zero, KILL cr);
 
@@ -11085,7 +11085,7 @@ instruct rep_stos_evex(rcx_RegL cnt, rdi_RegP base, regD tmp, kReg ktmp, rax_Reg
                        Universe dummy, rFlagsReg cr)
 %{
   predicate(!((ClearArrayNode*)n)->is_large() &&
-               UseAVX > 2 && VM_Version::supports_avx512vlbw() &&
+               UseAVX > 2 &&
                !n->in(2)->bottom_type()->is_long()->is_con());
   match(Set dummy (ClearArray cnt base));
   effect(USE_KILL cnt, USE_KILL base, TEMP tmp, TEMP ktmp, KILL zero, KILL cr);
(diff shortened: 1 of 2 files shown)


vnkozlov (approved):
Good.
What testing you did?

jatin-bhateja:
> Good.
> What testing you did?

No explicit test added since the code flow involves post allocation system initialization pattern which is very frequently exercised by almost all jtreg tests on KNL

vnkozlov:
I have to test it in our system before you push. I will let you know results.

vnkozlov:
tier1-3 testing passed.

src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -11085,7 +11085,7 @@ instruct rep_stos_evex(rcx_RegL cnt, rdi_RegP base, regD tmp, kReg ktmp, rax_Reg
                        Universe dummy, rFlagsReg cr)
 %{
   predicate(!((ClearArrayNode*)n)->is_large() &&
-               UseAVX > 2 && VM_Version::supports_avx512vlbw() &&
+               UseAVX > 2 &&
```
sviswa7, line 11088: VM_Version::supports_avx512vlbw() check is required otherwise MacroAssembler::clear_mem will assert. See the code snippet below. The assert is generated because the emovdqu with smaller than 512 bit vector width is not supported unless avx512vl() is available on the platform.
// Clearing constant sized memory using YMM/ZMM registers.
void MacroAssembler::clear_mem(Register base, int cnt, Register rtmp, XMMRegister xtmp, KRegister mask) {
  assert(UseAVX > 2 && VM_Version::supports_avx512vlbw(), "");

dean-long (commented):
I can't convince myself that this fix is correct.  The logic that controls when the mask register is used seems much more complicated that the predicate that guards it.

dean-long:
@jatin-bhateja I see your point.  With your change, the predicates are enough to prevent the assert.  But I noticed an existing issue for both the small and large cases.   The predicates do not prevent some cases where a mask register will be reserved when it isn't really needed.  For example when UseFastStosb is TRUE or UseXMMForObjInit is FALSE.  It seems like a good followup RFE.

dean-long (approved):

jatin-bhateja:
Hi @dean-long ,  yes extra opmask allocation will exist for non default control path, but its not a bug, and that's the penalty of packing mutiple flows in one pattern. We can do more cleanup in this flow as suggested.

PR #183, opened 2021-06-30
JDK-8225667: Clarify the behavior of System::gc w.r.t. reference processing

mlchung:
This spec clarification is a follow-up to [JDK-8224760](https://bugs.openjdk.java.net/browse/JDK-8224760?focusedCommentId=14268320&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14268320) w.r.t. reference processing.  Since there is no guarantee for any memory reclamation by the invocation of `System::gc`, the spec should also clarify that there is no guarantee in determining the change of reachability of any objects or any particular number of `Reference` objects be enqueued and cleared.

CSR:
   https://bugs.openjdk.java.net/browse/JDK-8269690

---------
### Progress

### Issue
 * [JDK-8225667](https://bugs.openjdk.java.net/browse/JDK-8225667): Clarify the behavior of System::gc w.r.t. reference processing

### Reviewers
 * [Roger Riggs](https://openjdk.java.net/census#rriggs) (@RogerRiggs - **Reviewer**)
 * [Kim Barrett](https://openjdk.java.net/census#kbarrett) (@kimbarrett - **Reviewer**)
 * [Thomas Schatzl](https://openjdk.java.net/census#tschatzl) (@tschatzl - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/183/head:pull/183` \
`$ git checkout pull/183`

`$ git checkout pull/183` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/183/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 183`

`$ git pr show -t 183`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/183.diff">https://git.openjdk.java.net/jdk17/pull/183.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/Runtime.java
+++ b/src/java.base/share/classes/java/lang/Runtime.java
@@ -656,9 +656,9 @@ public class Runtime {
      * number of unused objects, reclaim any particular amount of space, or
      * complete at any particular time, if at all, before the method returns or ever.
      * There is also no guarantee that this effort will determine
-     * the change of reachability in any particular number of objects
-     * or any particular number of {@link java.lang.ref.Reference Reference}
-     * objects be cleared and enqueued.
+     * the change of reachability in any particular number of objects,
+     * or that any particular number of {@link java.lang.ref.Reference Reference}
+     * objects will be cleared and enqueued.
      * <p>
      * The name {@code gc} stands for "garbage
      * collector". The Java Virtual Machine performs this recycling
--- a/src/java.base/share/classes/java/lang/Runtime.java
+++ b/src/java.base/share/classes/java/lang/Runtime.java
@@ -655,6 +655,10 @@ public class Runtime {
      * There is no guarantee that this effort will recycle any particular
      * number of unused objects, reclaim any particular amount of space, or
      * complete at any particular time, if at all, before the method returns or ever.
+     * There is also no guarantee that this effort will determine
+     * the change of reachability in any particular number of objects
+     * or any particular number of {@link java.lang.ref.Reference Reference}
+     * objects be cleared and enqueued.
      * <p>
      * The name {@code gc} stands for "garbage
      * collector". The Java Virtual Machine performs this recycling
--- a/src/java.base/share/classes/java/lang/System.java
+++ b/src/java.base/share/classes/java/lang/System.java
@@ -1862,9 +1862,9 @@ public final class System {
      * number of unused objects, reclaim any particular amount of space, or
      * complete at any particular time, if at all, before the method returns or ever.
      * There is also no guarantee that this effort will determine
-     * the change of reachability in any particular number of objects
-     * or any particular number of {@link java.lang.ref.Reference Reference}
-     * objects be cleared and enqueued.
+     * the change of reachability in any particular number of objects,
+     * or that any particular number of {@link java.lang.ref.Reference Reference}
+     * objects will be cleared and enqueued.
      *
      * <p>
      * The call {@code System.gc()} is effectively equivalent to the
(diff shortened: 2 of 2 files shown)


RogerRiggs (approved):

src/java.base/share/classes/java/lang/Runtime.java
```diff
@@ -655,6 +655,10 @@ public Process exec(String[] cmdarray, String[] envp, File dir)
      * There is no guarantee that this effort will recycle any particular
      * number of unused objects, reclaim any particular amount of space, or
      * complete at any particular time, if at all, before the method returns or ever.
+     * There is also no guarantee that this effort will determine
+     * the change of reachability in any particular number of objects
+     * or any particular number of {@link java.lang.ref.Reference Reference}
+     * objects be cleared and enqueued.
```
kimbarrett, line 661: perhaps
s/number of objects or any/number of objects, or that any/
s/objects be/objects will be/

src/java.base/share/classes/java/lang/System.java
```diff
@@ -1861,6 +1861,11 @@ public static void exit(int status) {
      * There is no guarantee that this effort will recycle any particular
      * number of unused objects, reclaim any particular amount of space, or
      * complete at any particular time, if at all, before the method returns or ever.
+     * There is also no guarantee that this effort will determine
+     * the change of reachability in any particular number of objects
+     * or any particular number of {@link java.lang.ref.Reference Reference}
+     * objects be cleared and enqueued.
```
kimbarrett, line 1867: Similar suggestion here as for Runtime.gc.

kimbarrett (changes requested):

src/java.base/share/classes/java/lang/Runtime.java
mlchung: That may be clearer.  Updated.

kimbarrett (approved):
Looks good.

tschatzl (approved):

RogerRiggs (approved):

src/java.base/share/classes/java/lang/Runtime.java
```diff
@@ -655,6 +655,10 @@ public Process exec(String[] cmdarray, String[] envp, File dir)
      * There is no guarantee that this effort will recycle any particular
      * number of unused objects, reclaim any particular amount of space, or
      * complete at any particular time, if at all, before the method returns or ever.
+     * There is also no guarantee that this effort will determine
+     * the change of reachability in any particular number of objects,
+     * or that any particular number of {@link java.lang.ref.Reference Reference}
+     * objects will be cleared and enqueued.
      * <p>
```
plevart, line 662: Hi Mandy,
I'm not a native speaker so this might be wrong thinking: The word "determine" feels to me like saying "cause". But running System.gc never actually causes the change of reachability (well it does, when the Reference object is cleared, the reachability of referent changes from Weakly/Softly/Phantom-reachable to unreachable, but I don't think you had that in mind. Or did you?). What would you say about using word "detect" instead? Please others, do say if my thinking is wrong...

plevart: Well, "determine" seems to have both meanings:
https://www.google.com/search?q=determine
So by the 2nd meaning, it is good.

plevart: ... and considering the "Schrödinger's cat", this interpretation is correct. You can't know whether the cat is dead or not until you "determine" it's death. This does not mean that you killed it though.

mlchung: I think "determine" sounds right to me.  "detect" is another possibility but does not apply well in "this effort" as the subject.   This is another way to explain it.

During the execution, there are a number of reachability decision points.  At each reachability decision point, the references are checked. GC determines if the reachability of the referent has changed to the value corresponding to the type of the reference (softly weakly, or phantom reachable), then it clears and enqueues the reference.

PR #185, opened 2021-06-30
8269697: JNI_GetPrimitiveArrayCritical() should not accept object array

zhengyu123:
GetPrimitiveArrayCritical() is supposed to only be used with primitive array types, but nothing prevents current implementation from accepting object arrays (please see attached test case in bug).

My purposed fix is not very friendly, it crashes JVM if a none primitive array is passed in, but I am sure what to expect in this scenario.

Specification people, please comment. Thanks!

---------
### Progress

### Issue
 * [JDK-8269697](https://bugs.openjdk.java.net/browse/JDK-8269697): JNI_GetPrimitiveArrayCritical() should not accept object array

### Reviewers
 * [Kim Barrett](https://openjdk.java.net/census#kbarrett) (@kimbarrett - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/185/head:pull/185` \
`$ git checkout pull/185`

`$ git checkout pull/185` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/185/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 185`

`$ git pr show -t 185`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/185.diff">https://git.openjdk.java.net/jdk17/pull/185.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/prims/jni.cpp
+++ b/src/hotspot/share/prims/jni.cpp
@@ -2822,13 +2822,8 @@ JNI_ENTRY(void*, jni_GetPrimitiveArrayCritical(JNIEnv *env, jarray array, jboole
     *isCopy = JNI_FALSE;
   }
   oop a = lock_gc_or_pin_object(thread, array);
-  assert(a->is_array(), "just checking");
-  BasicType type;
-  if (a->is_objArray()) {
-    type = T_OBJECT;
-  } else {
-    type = TypeArrayKlass::cast(a->klass())->element_type();
-  }
+  guarantee(a->is_typeArray(), "Primitive array only");
+  BasicType type = TypeArrayKlass::cast(a->klass())->element_type();
   void* ret = arrayOop(a)->base(type);
  HOTSPOT_JNI_GETPRIMITIVEARRAYCRITICAL_RETURN(ret);
   return ret;
--- a/src/hotspot/share/prims/jni.cpp
+++ b/src/hotspot/share/prims/jni.cpp
@@ -2822,7 +2822,7 @@ JNI_ENTRY(void*, jni_GetPrimitiveArrayCritical(JNIEnv *env, jarray array, jboole
     *isCopy = JNI_FALSE;
   }
   oop a = lock_gc_or_pin_object(thread, array);
-  guarantee(a->is_typeArray(), "Primitive array only");
+  assert(a->is_typeArray(), "Primitive array only");
   BasicType type = TypeArrayKlass::cast(a->klass())->element_type();
   void* ret = arrayOop(a)->base(type);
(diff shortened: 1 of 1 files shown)


src/hotspot/share/prims/jni.cpp
```diff
@@ -2822,13 +2822,8 @@ JNI_ENTRY(void*, jni_GetPrimitiveArrayCritical(JNIEnv *env, jarray array, jboole
...
   }
   oop a = lock_gc_or_pin_object(thread, array);
-  assert(a->is_array(), "just checking");
-  BasicType type;
-  if (a->is_objArray()) {
-    type = T_OBJECT;
-  } else {
-    type = TypeArrayKlass::cast(a->klass())->element_type();
-  }
+  guarantee(a->is_typeArray(), "Primitive array only");
```
kimbarrett, line 2825: I think an assert here is more usual.  This is JNI - we don't make guarantees (pun slightly intentional).  If you want error reporting, use -Xcheck:jni; the appropriate check and reporting is already there.

kimbarrett (changes requested):

kimbarrett (approved):
Looks good.

dholmes-ora (changes requested):
The change looks good.

We should add a testcase in runtime/jni/checked/ to ensure Xcheck:jni catches this.

Thanks,
David

shipilev:
Hold on. This introduces the behavioral change _for the worse_, right? More specifically, the current callers-with-object-array would now get `TypeArrayKlass::cast()`-ed to improper class with unknown consequences. _Hopefully_ it would be a JVM crash, but it might as well be incorrect execution. Old code would execute fine with most GCs (= the ones using GCLocker). Would it be saner to do a copy of object array if such a request is detected?

I believe it warrants a CSR discussion, and therefore postponement to JDK 18.

pliden:
The spec (https://docs.oracle.com/en/java/javase/16/docs/specs/jni/functions.html#getprimitivearraycritical-releaseprimitivearraycritical) says:

> Returns a pointer to the array elements, or NULL if the operation fails.

I interpret this as we're allowed to return NULL in this case. Operation failed because of invalid input.

shipilev:
Yes, returning `NULL` would be much saner here. That would alleviate my concern: the out-of-spec users would have a clear failure result. (It should be done before trying to lock/pin the object, of course).

zhengyu123:
The spec also says: 

> Therefore we need to check its return value against NULL for possible out of memory situations.

it suggests that returning NULL solely due to out of memory

zhengyu123:
Close this PR and move to 18 for further discussion.

PR #189, opened 2021-06-30
6766844: ByteArrayInputStream#read with a byte array of length 0 not consistent with InputStream when at EOF

bplb:
Modify the specification of `java.io.ByteArrayInputStream#read(byte[],int,int)` to indicate that `-1` is returned instead of `0` when the stream is at its end and the third parameter, `len`, is zero.

---------
### Progress

### Issue
 * [JDK-6766844](https://bugs.openjdk.java.net/browse/JDK-6766844): ByteArrayInputStream#read with a byte array of length 0 not consistent with InputStream when at EOF

### Reviewers
 * [Naoto Sato](https://openjdk.java.net/census#naoto) (@naotoj - **Reviewer**)
 * [Lance Andersen](https://openjdk.java.net/census#lancea) (@LanceAndersen - **Reviewer**)
 * [Joe Wang](https://openjdk.java.net/census#joehw) (@JoeWang-Java - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/189/head:pull/189` \
`$ git checkout pull/189`

`$ git checkout pull/189` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/189/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 189`

`$ git pr show -t 189`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/189.diff">https://git.openjdk.java.net/jdk17/pull/189.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/io/ByteArrayInputStream.java
+++ b/src/java.base/share/classes/java/io/ByteArrayInputStream.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1994, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1994, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -155,6 +155,10 @@ public class ByteArrayInputStream extends InputStream {
      * {@code buf[pos+k-1]} are copied into {@code b[off]} through
      * {@code b[off+k-1]} in the manner performed by {@code System.arraycopy}.
      * The value {@code k} is added into {@code pos} and {@code k} is returned.
+     * <p>
+     * Unlike the {@link InputStream#read(byte[],int,int) overridden method}
+     * of {@code InputStream}, this method returns {@code -1} instead of zero
+     * if the end of the stream has been reached and {@code len == 0}.
      * <p>
      * This {@code read} method cannot block.
      *
--- a/test/jdk/java/io/ByteArrayInputStream/ReadAllReadNTransferTo.java
+++ b/test/jdk/java/io/ByteArrayInputStream/ReadAllReadNTransferTo.java
@@ -51,10 +51,10 @@ public class ReadAllReadNTransferTo {
         ByteArrayInputStream bais = new ByteArrayInputStream(buf);
         bais.readAllBytes();
         if (bais.read(new byte[0]) != -1) {
-            throw new RuntimeException("read(byte[]) did not return 0");
+            throw new RuntimeException("read(byte[]) did not return -1");
         }
         if (bais.read(new byte[1], 0, 0) != -1) {
-            throw new RuntimeException("read(byte[],int,int) did not return 0");
+            throw new RuntimeException("read(byte[],int,int) did not return -1");
         }
 
         bais = new ByteArrayInputStream(buf, position, size);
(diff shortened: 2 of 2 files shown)


bplb:
This PR supersedes PR [4591](https://github.com/openjdk/jdk/pull/4591).

test/jdk/java/io/ByteArrayInputStream/ReadAllReadNTransferTo.java
```diff
@@ -48,8 +48,16 @@ public static void main(String... args) throws IOException {
...
 
-        ByteArrayInputStream bais =
-            new ByteArrayInputStream(buf, position, size);
+        ByteArrayInputStream bais = new ByteArrayInputStream(buf);
+        bais.readAllBytes();
+        if (bais.read(new byte[0]) != -1) {
+            throw new RuntimeException("read(byte[]) did not return 0");
+        }
+        if (bais.read(new byte[1], 0, 0) != -1) {
+            throw new RuntimeException("read(byte[],int,int) did not return 0");
```
naotoj, line 57: Should these exception messages be "did not return -1"?

bplb: Oh you are correct. Thanks!

naotoj (approved):

LanceAndersen (approved):

src/java.base/share/classes/java/io/ByteArrayInputStream.java
```diff
@@ -156,6 +156,10 @@ public synchronized int read() {
      * {@code b[off+k-1]} in the manner performed by {@code System.arraycopy}.
      * The value {@code k} is added into {@code pos} and {@code k} is returned.
      * <p>
+     * Unlike the {@link InputStream#read(byte[],int,int) overridden method}
+     * of {@code InputStream}, this method returns {@code -1} instead of zero
+     * if the end of the stream has been reached and {@code len == 0}.
```
JoeWang-Java, line 161: The statement "return -1 if the end of the stream has been reached and len == 0" gives an impression that it requires both conditions to be meet: end of the stream && len==0,  but the tests show -1 is expected if len == 0 without an attempt to read the stream. 

The overridden method stated that "If len is zero, then no bytes are read and 0 is returned", the above note looks like was meant for this statement since the overridden method also return -1 if the stream is at end of file.

bplb: Both conditions of the statement are intended to be met. If the stream is at its end then
```
        if (pos >= count) {
            return -1;
        }
```
will cause `-1` to be returned because at end of stream the condition `pos >= count` is met.

The overridden method always returns `0` if `len == 0`.:

```
    public int read(byte b[], int off, int len) throws IOException {
        Objects.checkFromIndexSize(off, len, b.length);
        if (len == 0) {
            return 0;
        }
```

JoeWang-Java: Right. I understand. The intention was, unlike the overridden method that returns 0 if len == 0 and -1 if at the end of the stream, this method returns -1 in both cases. A careful reader, after comparing both methods, would understand correctly that the difference is in the case of "len==0". I'm fine if you think this is good enough. Just a thought though that the statement could be interpreted as if both conditions need to be met at the same time (if "and" is taken as "&&", e.g. if (pos>==count && len==0) ). 

Something like the following might be clearer?
     * Unlike the {@link InputStream#read(byte[],int,int) overridden method}
     * of {@code InputStream} that returns {@code -1} if the end of the stream
     * has been reached and {@code 0} if {@code len == 0}, this method returns
     * {@code -1} in both cases.

Just my 2 cents.

bplb: Thanks for the suggestion. I am happy with it as is, but I'll hold off integrating it for now and rethink it later.

JoeWang-Java: Ok, good to know. Have a great weekend!

JoeWang-Java (approved):

src/java.base/share/classes/java/io/ByteArrayInputStream.java
bplb: I am inclined to leave the wording alone barring serious objections to the contrary.

PR #193, opened 2021-07-01
8269771: assert(tmp == _callprojs.fallthrough_catchproj) failed: allocation control projection

neliasso:
When I investigated this bug I found out it was already fixed by JDK-8268405, or at least - it didn't happen any more. Since I had a good reproducer I took a look at why it happened and found two issues.

The noteable thing about the test case is that there are still allocations that isn't used. The initalizeNode lead to HaltNodes. This looks like a temporary state while the graph is simplified.

1) In macro.cpp - process_users_of_allocation - there was an assert that was wrong. We must assert before the replace_node - otherwise tmp will always be NULL and fail. Why this hasn't happened before is a mystery.

2) memnode.cpp - MemBarNode::remove - returns without doing anything if there are not exactly 2 outs. In this case there is a InitalizeNode that leads to a HaltNode. The memory uses has been removed - so the memory projection from the InitalizeNode is already removed.

If you want to check if this patch fixes the issue - you need to sync to jdk-17+26 - or before the fix for JDK-8268405.

Please review,
Nils Eliasson

---------
### Progress

### Issue
 * [JDK-8269771](https://bugs.openjdk.java.net/browse/JDK-8269771): assert(tmp == _callprojs.fallthrough_catchproj) failed: allocation control projection

### Reviewers
 * [Rickard Bäckman](https://openjdk.java.net/census#rbackman) (@rickard - **Reviewer**) ⚠️ Review applies to 75336a9e5aeb9a6f62e29a4bfd0108a828aa2eb4

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/193/head:pull/193` \
`$ git checkout pull/193`

`$ git checkout pull/193` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/193/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 193`

`$ git pr show -t 193`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/193.diff">https://git.openjdk.java.net/jdk17/pull/193.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/macro.cpp
+++ b/src/hotspot/share/opto/macro.cpp
@@ -963,11 +963,12 @@ void PhaseMacroExpand::process_users_of_allocation(CallNode *alloc) {
         assert(init->outcnt() <= 2, "only a control and memory projection expected");
         Node *ctrl_proj = init->proj_out_or_null(TypeFunc::Control);
         if (ctrl_proj != NULL) {
-#ifdef ASSERT
-          Node* tmp = init->in(TypeFunc::Control);
-          assert(tmp == _callprojs.fallthrough_catchproj, "allocation control projection");
-#endif
           _igvn.replace_node(ctrl_proj, init->in(TypeFunc::Control));
+#ifdef ASSERT
+          // If the InitializeNode has no memory out, it will die, and tmp will become NULL
+          Node* tmp = init->in(TypeFunc::Control);
+          assert(tmp == NULL || tmp == _callprojs.fallthrough_catchproj, "allocation control projection");
+#endif
         }
         Node *mem_proj = init->proj_out_or_null(TypeFunc::Memory);
         if (mem_proj != NULL) {
--- a/src/hotspot/share/opto/macro.cpp
+++ b/src/hotspot/share/opto/macro.cpp
@@ -963,11 +963,11 @@ void PhaseMacroExpand::process_users_of_allocation(CallNode *alloc) {
         assert(init->outcnt() <= 2, "only a control and memory projection expected");
         Node *ctrl_proj = init->proj_out_or_null(TypeFunc::Control);
         if (ctrl_proj != NULL) {
-          _igvn.replace_node(ctrl_proj, init->in(TypeFunc::Control));
 #ifdef ASSERT
           Node* tmp = init->in(TypeFunc::Control);
           assert(tmp == _callprojs.fallthrough_catchproj, "allocation control projection");
 #endif
+          _igvn.replace_node(ctrl_proj, init->in(TypeFunc::Control));
         }
         Node *mem_proj = init->proj_out_or_null(TypeFunc::Memory);
         if (mem_proj != NULL) {
--- a/src/hotspot/share/opto/memnode.cpp
+++ b/src/hotspot/share/opto/memnode.cpp
@@ -3313,7 +3313,8 @@ MemBarNode* MemBarNode::make(Compile* C, int opcode, int atp, Node* pn) {
 
 void MemBarNode::remove(PhaseIterGVN *igvn) {
   if (outcnt() != 2) {
-    return;
+    assert(Opcode() == Op_Initialize, "Only seen when there are no use of init memory");
+    assert(outcnt() == 1, "Only control then");
   }
   if (trailing_store() || trailing_load_store()) {
     MemBarNode* leading = leading_membar();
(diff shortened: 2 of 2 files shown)


rickard (approved):

rickard (approved):

vnkozlov (changes requested):
I suggest to create new **open** bug and close current as dup.

I also don't get why your change in macro.cpp fixed the issue. `fallthrough_catchproj` is local cached value - it should not be updated with call to `replace_node()`.

Can you tell more what happened here?

neliasso:
> I suggest to create new **open** bug and close current as dup.

This fixes the original reported bug - so that we run fine even without JDK-8268405 (which probably doesn't fix it - just hides it.)

> I also don't get why your change in macro.cpp fixed the issue. `fallthrough_catchproj` is local cached value - it should not be updated with call to `replace_node()`.
> 
> Can you tell more what happened here?

fallthrough_catchproj is chached. It's the tmp that will be NULL if the assert is after the replace. The Init-node is being removed. Maybe I should just remove the entire assert - it's a bit unnecessary.

vnkozlov:
> > I suggest to create new **open** bug and close current as dup.
> 
> This fixes the original reported bug - so that we run fine even without JDK-8268405 (which probably doesn't fix it - just hides it.)

I mean this bug 8269641 is marked as `Confidential`. Nobody can see it, even GitHub - see `Issue`. You need to create new bug which does not have our confidential information for review in open GitHub.

> 
> > I also don't get why your change in macro.cpp fixed the issue. `fallthrough_catchproj` is local cached value - it should not be updated with call to `replace_node()`.
> > Can you tell more what happened here?
> 
> fallthrough_catchproj is chached. It's the tmp that will be NULL if the assert is after the replace. The Init-node is being removed. Maybe I should just remove the entire assert - it's a bit unnecessary.

It looks like `InitalizeNode` has only control projection output and when it is replaced it also removes `InitalizeNode` as dead node too. In such case you just need to relax assert by checking `tmp == NULL ||` and adding comment that `InitalizeNode` can become dead when it has only one projection.

neliasso:
> > > I suggest to create new **open** bug and close current as dup.
> > 
> > 
> > This fixes the original reported bug - so that we run fine even without JDK-8268405 (which probably doesn't fix it - just hides it.)
> 
> I mean this bug 8269641 is marked as `Confidential`. Nobody can see it, even GitHub - see `Issue`. You need to create new bug which does not have our confidential information for review in open GitHub.
> 
> > > I also don't get why your change in macro.cpp fixed the issue. `fallthrough_catchproj` is local cached value - it should not be updated with call to `replace_node()`.
> > > Can you tell more what happened here?
> > 
> > 
> > fallthrough_catchproj is chached. It's the tmp that will be NULL if the assert is after the replace. The Init-node is being removed. Maybe I should just remove the entire assert - it's a bit unnecessary.
> 
> It looks like `InitalizeNode` has only control projection output and when it is replaced it also removes `InitalizeNode` as dead node too. In such case you just need to relax assert by checking `tmp == NULL ||` and adding comment that `InitalizeNode` can become dead when it has only one projection.

Changed bug, and added check with comment.

vnkozlov (approved):
Good.

navyxliu:
hi, @vnkozlov , 

Are we handling code like this? 
```
 \
-o----------
| Initializer |
-----------
  \
   o------
   | Halt   |  
   ---------
```

> It looks like InitalizeNode has only control projection output and when it is replaced it also removes InitalizeNode as dead node too. 

Yes, ` replace_node(ctrl_proj, init->in(TypeFunc::Control));` calls `subsume_node(halt, init->in(TypeFunc::Control))`, and it will empty halt's outputs, but can we say that init->in(TypeFunc::Control) is NULL after replace_node()?  
Initializer is a dead node after then, but I don't see its in(0) become NULL.

vnkozlov:
hi @navyxliu
`ctrl_proj` and `init->in(TypeFunc::Control)` (which is Allocation's control out `Proj` node) are separate `Proj` nodes. After calling  `replace_node()` `Halt` node will be attached to Allocation's control out `Proj` node.
 No users of `ctrl_proj` left after control edge of Halt node is replaced. `remove_dead_node(old)` is called for it: [phaseX.cpp#L1488](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/opto/phaseX.cpp#L1488)
Which calls `remove_globally_dead_node(dead)` and in it input edges of `ctrl_proj` are replaced with NULL:
[phaseX.cpp#L1383](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/opto/phaseX.cpp#L1383) 
and it checks if its input (which is `Initalize` node) has any other output (users). If input node does not have any - it is also dead and placed on stack for removal. As result `Initalize` node have its inputs also replaced with NULL.

navyxliu:
Thank you for the explanation!  I missed `remove_dead_node( old )` in subsume_node. now I see why initilizer->in(0) will become NULL after ctrl_proj is replaced. 

This PR looks good to me.

PR #194, opened 2021-07-01
8269146: Missing unreported constraints on pattern and other case label combination

lahodaj:
There is a number of constraints on the case label forms - patterns cannot be combined with patterns, etc. Currently, the checks are implemented by checking if there are clashing bindings. But this misses some of the cases which should lead to errors. This patch proposes to implement explicit checks on the structure of the case labels in `Attr.handleSwitch`, rather than relying on clashing bindings.

---------
### Progress

### Issues
 * [JDK-8269146](https://bugs.openjdk.java.net/browse/JDK-8269146): Missing unreported constraints on pattern and other case label combination
 * [JDK-8269301](https://bugs.openjdk.java.net/browse/JDK-8269301): Switch statement with a pattern, constant and default label elements crash javac

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/194/head:pull/194` \
`$ git checkout pull/194`

`$ git checkout pull/194` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/194/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 194`

`$ git pr show -t 194`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/194.diff">https://git.openjdk.java.net/jdk17/pull/194.diff</a>

</details>


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1681,10 +1681,18 @@ private void handleSwitch(JCTree switchTree,
             boolean hasDefault = false;           // Is there a default label?
             boolean hasTotalPattern = false;      // Is there a total pattern?
             boolean hasNullPattern = false;       // Is there a null pattern?
+            boolean wasConstant = false;          // Seen a constant in the same case label
```
mcimadamore, line 1684: I'd suggest to move the `wasXYZ` variables inside the case label loop - since they are not meant to be variables that are global to the entire switch construct?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1806,11 +1825,17 @@ private void handleSwitch(JCTree switchTree,
                 addVars(c.stats, switchEnv.info.scope);
 
                 boolean completesNormally = c.caseKind == CaseTree.CaseKind.STATEMENT ? flow.aliveAfter(caseEnv, c, make) : false;
+
+                if (c.stats.nonEmpty()) {
```
mcimadamore, line 1829: This initialization should probably go at the beginning of the for loop block?

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
```diff
@@ -1702,72 +1710,84 @@ private void handleSwitch(JCTree switchTree,
...
                         }
                         hasDefault = true;
+                        wasDefault = true;
                         matchBindings = MatchBindingsComputer.EMPTY;
                     } else {
-                        if (prevCompletedNormally) {
+                        boolean isTypePattern = pat.hasTag(BINDINGPATTERN);
+                        if (wasPattern || wasConstant || wasDefault ||
+                            (wasNullPattern && (!isTypePattern || wasNonEmptyFallThrough))) {
                             log.error(pat.pos(), Errors.FlowsThroughToPattern);
```
mcimadamore, line 1787: Is this a good error message for the condition? E.g. it seems like here we're mostly complaining about some illegal mix of constant, default and pattern labels - which seems more something to do with well-formedness than with fallthrough?

mcimadamore (commented):
I wonder if all the `wasDefault` etc tests, to check compatibility between different kind of labels couldn't be moved when at parse time. The code in attr is already quite convoluted, and it seems to be doing many things at once - if there are checks which depend on flow analysis, of course these should remain in Attr - but there seems to be a lot of structural checks which do not really depend on attribution which could be caught a lot earlier, and probably giving better error messages too.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Attr.java
lahodaj: I agree it would be much better to have the flags closer, but the flags sadly generally apply across case clauses. The conditions don't allow `case <type pattern>, default:` as well as `case <type pattern>: /*no statements*/ case default:`, because both are `SwitchLabel`s. So we need to keep information across `case` clauses if the case has an empty list of statements, and in some cases even if the list of statements is not empty, but there is a fall-through to the following case (which is to prevent cases like `case null: System.err.println(); /*no break*/ case Integer i:`).

mcimadamore: OK, thanks for the clarification, I think I now get what the code is trying to do and the fact that multiple `cases` (even if separated by commas) are really dealt with separately by the logic, so you need to keep these flags in scope for longer.

I still think that some of these checks could be perhaps moved closer to parser - and then in Attr we could check stuff that depends on types (e.g. dominance), or stuff that depends on control flow (e.g. cannot fall through to a pattern).

I understand that the code was like this (e.g. duplicate labels, or duplicate defaults has always been caught in Attr) - but as the number of checks grows larger, it feels like keep adding checks in Attr doesn't scale very well.

mcimadamore (approved):
Approving for now, perhaps with a view to investigate ways to split the checks at a later point.

lahodaj:
I've updated the patch so that the new (/structural) tests are moved from Attr to Check. To me, looks better. What do you think?
Thanks!

mcimadamore (approved):
Looks good - I agree the code looks simpler, and there's a bit more separation of concerns

PR #198, opened 2021-07-01
8269281: java/foreign/TestUpcall.java times out 

mcimadamore:
The previous fix to this test which used removed the `VerifyDependency` flag worked well, but we're still observing rare timeouts. Looking at the test reports, it seems likely that slightly increasing the timeout would do the trick.

---------
### Progress

### Issue
 * [JDK-8269281](https://bugs.openjdk.java.net/browse/JDK-8269281): java/foreign/TestUpcall.java times out

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/198/head:pull/198` \
`$ git checkout pull/198`

`$ git checkout pull/198` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/198/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 198`

`$ git pr show -t 198`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/198.diff">https://git.openjdk.java.net/jdk17/pull/198.diff</a>

</details>

### DIFF
--- a/test/jdk/java/foreign/TestUpcall.java
+++ b/test/jdk/java/foreign/TestUpcall.java
@@ -28,7 +28,7 @@
  * @modules jdk.incubator.foreign/jdk.internal.foreign
  * @build NativeTestHelper CallGeneratorHelper TestUpcall
  *
- * @run testng/othervm -XX:+IgnoreUnrecognizedVMOptions -XX:-VerifyDependencies
+ * @run testng/othervm/timeout=240 -XX:+IgnoreUnrecognizedVMOptions -XX:-VerifyDependencies
  *   --enable-native-access=ALL-UNNAMED
  *   TestUpcall
  */


dholmes-ora:
Hi Maurizio,

Sorry but I don't see why you think increasing the timeout will help here. From what I can see a normal successful run takes less than a minute, or just over. I see a couple of outliers at 4 and 6 minutes, then you have a timeout after 10 minutes - that is 10x slower than expected. I think something unexpected is happening here to cause it to take so long.

David

mcimadamore:
After an offline discussion we agreed to leave the test unchanged at the moment. It is not immediately clear if the problem is the test (which runs in ~40 seconds) or with specific machines, as we're seeing some very wild spikes. The test uses some logic to prod the GC (so that cleaners are run and off-heap resources are released) - but the same logic is also present in the dual TestDowncall which has never timed out.

PR #200, opened 2021-07-02
8269668: [aarch64] java.library.path not including /usr/lib64

jerboaa:
This fix simplifies the `if` condition to use a default 64-bit library path including default setting when `_LP64` is defined (os has 64-bitness). This should be safe IMHO and seems better than explicitly enumerating all 64 bit Linux architectures.

This issue affects older releases and I'd like to get this fix into JDK 17. Thoughts?

---------
### Progress

### Issue
 * [JDK-8269668](https://bugs.openjdk.java.net/browse/JDK-8269668): [aarch64] java.library.path not including /usr/lib64

### Reviewers
 * [Andrew Haley](https://openjdk.java.net/census#aph) (@theRealAph - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/200/head:pull/200` \
`$ git checkout pull/200`

`$ git checkout pull/200` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/200/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 200`

`$ git pr show -t 200`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/200.diff">https://git.openjdk.java.net/jdk17/pull/200.diff</a>

</details>

### DIFF
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -412,7 +412,7 @@ void os::init_system_properties_values() {
   //        ...
   //        7: The default directories, normally /lib and /usr/lib.
 #ifndef OVERRIDE_LIBPATH
-  #if defined(AMD64) || (defined(_LP64) && defined(SPARC)) || defined(PPC64) || defined(S390)
+  #if defined(_LP64)
     #define DEFAULT_LIBPATH "/usr/lib64:/lib64:/lib:/usr/lib"
   #else
     #define DEFAULT_LIBPATH "/lib:/usr/lib"


theRealAph:
I'm trying to think of ways this might break.

I'm not sure about "should be safe". I'm not an expert on Debian and
Ubuntu, but AFAIK /usr/lib is the installed system's default arch,
with multi-arch stuff like /usr/lib/i386-linux-gnu, etc. On such
systems I believe that /usr/lib64 is often symlinked to point to
/usr/lib, but it might not be: it wouldn't be completely insane to
have some other arrangement.  It could be argued that if the Debians
want to violate the FHS it's on them to patch around it, but this
patch is a change of behaviour, so it may not be entirely safe for
back-porting.

jerboaa:
> I'm trying to think of ways this might break.

Good, thanks! I came up empty where adding 64-bit paths in addition (and earlier on the lookup path) might actually be a problem.

> I'm not sure about "should be safe".

:) That comment was meant to be seen in the context of JDK 17/18. I agree for backporting I'd rather use the alternative patch as suggested in https://github.com/openjdk/jdk/pull/4657. Now, in the context of JDK 17 (and JDK 18) which has the SPARC port removed, it seems that simplification does make sense. It might not be 100% safe, I agree.

> I'm not an expert on Debian and
> Ubuntu, but AFAIK /usr/lib is the installed system's default arch,
> with multi-arch stuff like /usr/lib/i386-linux-gnu, etc. On such
> systems I believe that /usr/lib64 is often symlinked to point to
> /usr/lib, but it might not be: it wouldn't be completely insane to
> have some other arrangement. It could be argued that if the Debians
> want to violate the FHS it's on them to patch around it, but this
> patch is a change of behaviour, so it may not be entirely safe for
> back-porting.

Yes, agreed. Not safe for backporting.

dholmes-ora:
I'm a bit confused. On all 64-bit Linux platforms we set the same default paths, but the change for Aarch64 was overlooked. So unless these distributions set different paths for different architectures surely we want the same behaviour on all Linux platforms?

theRealAph:
> I'm a bit confused. On all 64-bit Linux platforms we set the same default paths, but the change for Aarch64 was overlooked. So unless these distributions set different paths for different architectures surely we want the same behaviour on all Linux platforms?

We do now, sure, except for platforms that use "/lib" for 64-bit libraries. These days "/lib64" is used on all FHS-compliant systems, but Debian is not FHS-compliant and some Linux systems date from before "/lib64" was decided on

jerboaa:
> > I'm a bit confused. On all 64-bit Linux platforms we set the same default paths, but the change for Aarch64 was overlooked. So unless these distributions set different paths for different architectures surely we want the same behaviour on all Linux platforms?
> 
> We do now, sure, except for platforms that use "/lib" for 64-bit libraries. These days "/lib64" is used on all FHS-compliant systems, but Debian is not FHS-compliant and some Linux systems date from before "/lib64" was decided on

Considering a platform that has a 64-bit library `foo` in `/lib` (over `/lib64`), then a call to `System.loadLibrary("foo")` would continue to work (as this patch just adds more default paths). It would only break on those systems that have `/lib64/libfoo.so` *and* `/lib/libfoo.so` where the former is actually different. I wonder how likely this is. For ppc64, s390x, amd64 we already have the 64-bit including default paths. In addition, since JDK-8214332 - JDK 12 -, there is `--with-jni-libpath` configure option to override all that.

All things considered, the post-patch fallback default library path seems reasonable.

theRealAph:
> Considering a platform that has a 64-bit library `foo` in `/lib` (over `/lib64`), then a call to `System.loadLibrary("foo")` would continue to work (as this patch just adds more default paths). It would only break on those systems that have `/lib64/libfoo.so` _and_ `/lib/libfoo.so` where the former is actually different. I wonder how likely this is. For ppc64, s390x, amd64 we already have the 64-bit including default paths. In addition, since JDK-8214332 - JDK 12 -, there is `--with-jni-libpath` configure option to override all that.
> 
> All things considered, the post-patch fallback default library path seems reasonable.

Make it `/lib:/lib64` for backports and you don't have to worry about that possibility.

All this nonsense could be avoided if we used ldconfig rather than hard-coded strings, but we are where we are.

jerboaa:
> Make it `/lib:/lib64` for backports and you don't have to worry about that possibility.

Yes, I'll be sure to not go straight to backporting this patch verbatim. A more conservative fix would be needed there.

jerboaa:
OK for JDK 17?

theRealAph (approved):

dholmes-ora (approved):
My take is that this should not be an architecture specific setting, so if it is desirable and works on existing x86_64, PPC64 and S390 platforms then it should also be desirable and work for Aarch64. If some distribution differs for a given architecture then that distribution can rebuild with OVERRIDE_LIBPATH as appropriate.

Thanks,
David

jerboaa:
Thanks for the reviews!

PR #202, opened 2021-07-02
8268859: jshell throws exception while parsing illegal "case true"

lahodaj:
Currently, broken case labels like `case true t` parse as patterns, whose type is `true`, which then fails later during annotation handling. This patch is trying to improve (and fix) the pattern/expression disambiguation again, making it more explicit than before. With this patch, only labels that begin with `<type> <name>` should be considered patterns.

The recently added no-preview part of `SwitchErrors` is moved to `PatternErrorRecovery.java`, because that indeed turned out to be difficult to maintain.

---------
### Progress

### Issue
 * [JDK-8268859](https://bugs.openjdk.java.net/browse/JDK-8268859): jshell throws exception while parsing illegal "case true"

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/202/head:pull/202` \
`$ git checkout pull/202`

`$ git checkout pull/202` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/202/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 202`

`$ git pr show -t 202`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/202.diff">https://git.openjdk.java.net/jdk17/pull/202.diff</a>

</details>


mcimadamore (approved):
Parser changes look good - splitting lambda analysis from pattern analysis and finding commonalities (e.g. skipping annotations) seems like a wise move.

PR #208, opened 2021-07-03
8269820: C2 PhaseIdealLoop::do_unroll get wrong opaque node

huishi-hs:
In PhaseIdealLoop::is_canonical_loop_entry check, opaque node can be at either input of compare node, after is_canonical_loop_entry check, opaq node is getting only from compare node's in(2). This might get wrong oqaque node and cause crash/assertion.  Detailed crash and analysis is in JBS.

Fix: Adding method PhaseIdealLoop::get_opaque_from_cmp, getting opaque node from cmp node's either input.

Test: Linux X64 tier1/2/3 release/fastdebug no regression.

After this fix, BoolNode::Ideal might remove its code forbidding swapping compare node's input node order when second one is opaque1 node.

---------
### Progress

### Error
&nbsp;⚠️ This PR only contains changes already present in the target

### Issue
 * [JDK-8269820](https://bugs.openjdk.java.net/browse/JDK-8269820): C2 PhaseIdealLoop::do_unroll get wrong opaque node

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/208/head:pull/208` \
`$ git checkout pull/208`

`$ git checkout pull/208` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/208/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 208`

`$ git pr show -t 208`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/208.diff">https://git.openjdk.java.net/jdk17/pull/208.diff</a>

</details>


rwestrel:
Have you considered preventing split-thru-phi with an opaque input? That would be a better fix IMO.

huishi-hs:
> Have you considered preventing split-thru-phi with an opaque input? That would be a better fix IMO.

Thanks for your comment! I'll check other fix and back to you.

huishi-hs:
@rwestrel sorry for late update, I did detail checks about how opaque1 get cloned in loop optimization. Preventing split-thru-phi with an opaque input doesn't work in my investigation.

image
IdealGraph in http://cr.openjdk.java.net/~hshi/8269820/loop_unroll.xml

1. Opaque1 node is not split in split_thru_phi, it is split when split if in split_up.
2. Entire split is triggered by Split if on if node 415
3. Split starts at region node 210 and phi node 213, PhaseIdealLoop::split_up from 213->492->497(recursive to 516)->516->499->458
4. If disable split 458 in PhaseIdealLoop::split_up, it will crash with other assertions, because "do_split_if" expects "have no instructions in the block containing the IF".

rwestrel:
@huishi-hs Thanks for exploring this fix. I still think blocking split thru phi is the way to go. Otherwise there's a risk that the Opaque1 ends up behind a Phi which would break the pattern entirely. This would require something like this:

https://github.com/rwestrel/jdk17/commit/e6d6ebecba478d0a080954494b927d4243ca2ee4

huishi-hs:
> @huishi-hs Thanks for exploring this fix. I still think blocking split thru phi is the way to go. Otherwise there's a risk that the Opaque1 ends up behind a Phi which would break the pattern entirely. This would require something like this:
> 
> [rwestrel@e6d6ebe](https://github.com/rwestrel/jdk17/commit/e6d6ebecba478d0a080954494b927d4243ca2ee4)

Thanks!  Checking if opaque1 node will be splited in split_if  looks heavy and this might affect current optimization(might have unexpected degradation). I will do some compile time estimate first.

huishi-hs:
@rwestrel 
Testing https://github.com/rwestrel/jdk17/commit/e6d6ebecba478d0a080954494b927d4243ca2ee4, testing with -Xcomp -XX:-TieredCompilation (adding elapsedTimer for PhaseIdealLoop::split_if_with_blocks_post and merge_point_safe),  5% - 10% overhead for this extra check in entier split_if_with_blocks_post.  
```
[Total extra check            , 0.0000035 secs]
[Total after can_split_if time, 0.0000646 secs]
[Total extra check            , 0.0000032 secs]
[Total after can_split_if time, 0.0000649 secs]
[Total extra check            , 0.0000023 secs]
[Total after can_split_if time, 0.0000604 secs]
[Total extra check            , 0.0000023 secs]
[Total after can_split_if time, 0.0000214 secs]
[Total extra check            , 0.0000020 secs]
[Total after can_split_if time, 0.0000198 secs]
[Total after can_split_if time, 0.0000010 secs]
[Total extra check            , 0.0000030 secs]
[Total after can_split_if time, 0.0000383 secs]
[Total extra check            , 0.0000026 secs]
[Total after can_split_if time, 0.0000682 secs]

```
This also disables split if might lost some optimization and might have unexpected degradation.  Could you please review new fix? New fix is disabling swapping compare node's input in BoolNode.Ideal, when its second input is phi of opaque1 node.

Before transformation, opaque1 is performed on comve.
image

After transformation, canonical loop limit is constant.
image

huishi-hs:
Close this as conflict with fix for https://bugs.openjdk.java.net/browse/JDK-8269752

PR #215, opened 2021-07-05
8269879: [PPC64] C2: Math.rint intrinsic uses wrong rounding mode

TheRealMDoerr:
We need to replace the frin and xvrdpi instructions which use a wrong rounding mode. xvrdpic is available with Power7. We can simply switch off RoundDoubleMode intrinsics for older processors (which are outdated).
Note that xvrdpic requires RN=0b00 to use the correct rounding mode. These 2 bits are 0 by default (see "Registers Specified during Process Initialization" in [1]) and are treated as "Limited-Access Bits" (see "2.2.1.2. Limited-Access Bits" in [1] for preservation rules).

[1] "64-Bit ELF V2 ABI Specification" http://cdn.openpowerfoundation.org/wp-content/uploads/resources/leabi/leabi-20170510.pdf

---------
### Progress

### Issue
 * [JDK-8269879](https://bugs.openjdk.java.net/browse/JDK-8269879): [PPC64] C2: Math.rint intrinsic uses wrong rounding mode

### Reviewers
 * [Lutz Schmidt](https://openjdk.java.net/census#lucy) (@RealLucy - **Reviewer**)
 * [Goetz Lindenmaier](https://openjdk.java.net/census#goetz) (@GoeLin - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/215/head:pull/215` \
`$ git checkout pull/215`

`$ git checkout pull/215` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/215/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 215`

`$ git pr show -t 215`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/215.diff">https://git.openjdk.java.net/jdk17/pull/215.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/ppc/assembler_ppc.hpp
+++ b/src/hotspot/cpu/ppc/assembler_ppc.hpp
@@ -576,6 +576,7 @@ class Assembler : public AbstractAssembler {
     XVNMSUBASP_OPCODE=(60u<< OPCODE_SHIFT |  209u << 3),
     XVNMSUBADP_OPCODE=(60u<< OPCODE_SHIFT |  241u << 3),
     XVRDPI_OPCODE  = (60u << OPCODE_SHIFT |  201u << 2),
+    XVRDPIC_OPCODE = (60u << OPCODE_SHIFT |  235u << 2),
     XVRDPIM_OPCODE = (60u << OPCODE_SHIFT |  249u << 2),
     XVRDPIP_OPCODE = (60u << OPCODE_SHIFT |  233u << 2),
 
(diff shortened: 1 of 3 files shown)


RealLucy (changes requested):
The changes look good so far.

You are mentioning that the now used instruction is available from Power7 and that for older versions of the Power spec, the intrinsic should be disabled. I am missing such a change, probably in vm_version_ppc.cpp.

Furthermore, you mention that the instruction needs a special (non-default) setting of the rounding mode control flags. How can such flag setting be assured? 
Thanks, Lutz

TheRealMDoerr:
> The changes look good so far.

Thanks!

> You are mentioning that the now used instruction is available from Power7 and that for older versions of the Power spec, the intrinsic should be disabled. I am missing such a change, probably in vm_version_ppc.cpp.

See change in Matcher::match_rule_supported(int opcode).

> Furthermore, you mention that the instruction needs a special (non-default) setting of the rounding mode control flags. How can such flag setting be assured?

No, it's the default setting (see referenced ABI chapters).

RealLucy (approved):
Thanks for your clarification! My "prefabricated" opinion blurred the sights on reality. All fine!

GoeLin (approved):
LGTM

TheRealMDoerr:
Thanks for the reviews!
/integrate

PR #216, opened 2021-07-06
8269865: Async UL needs to handle ERANGE on exceeding SEM_VALUE_MAX

navyxliu:
This patch solved the sempahore overflow issue with errno ERANGE or EOVERFLOW.
Previously, we have asymmetric p/v operations for semaphore _sem. Each iteration
only decrements _sem 1 but dequeues N messages. If logging threads keep preempting
async logging thread, it may cause the value of _sem accumulates until overflow! 

We decide to ditch counting semaphore.  os::PlatformMonitor can be used as low-level 
combination of Mutex/CondVar.  It may work if we simply ignore Semaphore::signal failure 
due to value overflow, but we have to pollute Semaphore interface.  Also it seems that MacOS
semaphore which is non-posix doesn't fail in this case.

---------
### Progress

### Issue
 * [JDK-8269865](https://bugs.openjdk.java.net/browse/JDK-8269865): Async UL needs to handle ERANGE on exceeding SEM_VALUE_MAX

### Reviewers
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to 0816eb2dd2aef357b33aa4e7eb8038acf1147a8f
 * [Patricio Chilano Mateo](https://openjdk.java.net/census#pchilanomate) (@pchilano - Committer) ⚠️ Review applies to 0816eb2dd2aef357b33aa4e7eb8038acf1147a8f

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/216/head:pull/216` \
`$ git checkout pull/216`

`$ git checkout pull/216` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/216/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 216`

`$ git pr show -t 216`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/216.diff">https://git.openjdk.java.net/jdk17/pull/216.diff</a>

</details>

### DIFF
--- a/src/hotspot/os/windows/semaphore_windows.cpp
+++ b/src/hotspot/os/windows/semaphore_windows.cpp
@@ -39,7 +39,7 @@ WindowsSemaphore::~WindowsSemaphore() {
   ::CloseHandle(_semaphore);
 }
 
-void WindowsSemaphore::signal(uint count) {
+void WindowsSemaphore::signal(uint count, bool ignore_overflow) {
   if (count > 0) {
     BOOL ret = ::ReleaseSemaphore(_semaphore, count, NULL);
 
--- a/src/hotspot/os/windows/semaphore_windows.cpp
+++ b/src/hotspot/os/windows/semaphore_windows.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -39,21 +39,12 @@ WindowsSemaphore::~WindowsSemaphore() {
   ::CloseHandle(_semaphore);
 }
 
-bool WindowsSemaphore::signal(uint count, bool ignore_overflow) {
-  bool succeed = true;
-
+void WindowsSemaphore::signal(uint count) {
   if (count > 0) {
     BOOL ret = ::ReleaseSemaphore(_semaphore, count, NULL);
-    DWORD err = GetLastError();
 
-    if (ignore_overflow && ret == 0) {
-      assert(err == ERROR_TOO_MANY_POSTS, "ReleaseSemaphore failed due to overflow");
-      succeed = false;
-    } else {
-      assert(ret != 0, "ReleaseSemaphore failed with error code: %lu", err);
-    }
+    assert(ret != 0, "ReleaseSemaphore failed with error code: %lu", GetLastError());
   }
-  return succeed;
 }
 
 void WindowsSemaphore::wait() {
--- a/src/hotspot/os/windows/semaphore_windows.cpp
+++ b/src/hotspot/os/windows/semaphore_windows.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -42,8 +42,9 @@ WindowsSemaphore::~WindowsSemaphore() {
 void WindowsSemaphore::signal(uint count, bool ignore_overflow) {
   if (count > 0) {
     BOOL ret = ::ReleaseSemaphore(_semaphore, count, NULL);
-
-    assert(ret != 0, "ReleaseSemaphore failed with error code: %lu", GetLastError());
+    DWORD err = GetLastError();
+    assert(ret != 0 || (ignore_overflow && ERROR_TOO_MANY_POSTS == err),
+          "ReleaseSemaphore failed with error code: %lu", err);
   }
 }
 
--- a/src/hotspot/os/windows/semaphore_windows.cpp
+++ b/src/hotspot/os/windows/semaphore_windows.cpp
@@ -39,13 +39,21 @@ WindowsSemaphore::~WindowsSemaphore() {
   ::CloseHandle(_semaphore);
 }
 
-void WindowsSemaphore::signal(uint count, bool ignore_overflow) {
+bool WindowsSemaphore::signal(uint count, bool ignore_overflow) {
+  bool succeed = true;
+
   if (count > 0) {
     BOOL ret = ::ReleaseSemaphore(_semaphore, count, NULL);
     DWORD err = GetLastError();
-    assert(ret != 0 || (ignore_overflow && ERROR_TOO_MANY_POSTS == err),
-          "ReleaseSemaphore failed with error code: %lu", err);
+
+    if (ignore_overflow && ret == 0) {
+      assert(err == ERROR_TOO_MANY_POSTS, "ReleaseSemaphore failed due to overflow");
+      succeed = false;
+    } else {
+      assert(ret != 0, "ReleaseSemaphore failed with error code: %lu", err);
+    }
   }
+  return succeed;
 }
 
 void WindowsSemaphore::wait() {
--- a/src/hotspot/share/logging/logAsyncWriter.cpp
+++ b/src/hotspot/share/logging/logAsyncWriter.cpp
@@ -51,7 +51,7 @@ void AsyncLogWriter::enqueue_locked(const AsyncLogMessage& msg) {
   }
 
   _buffer.push_back(msg);
-  _sem.signal();
+  _sem.signal_overflow();
 }
 
 void AsyncLogWriter::enqueue(LogFileOutput& output, const LogDecorations& decorations, const char* msg) {
@@ -198,7 +198,7 @@ void AsyncLogWriter::flush() {
 
       // Push directly in-case we are at logical max capacity, as this must not get dropped.
       _instance->_buffer.push_back(token);
-      _instance->_sem.signal();
+      _instance->_sem.signal_overflow();
     }
 
     _instance->_flush_sem.wait();
--- a/src/hotspot/share/logging/logAsyncWriter.cpp
+++ b/src/hotspot/share/logging/logAsyncWriter.cpp
@@ -27,6 +27,7 @@
 #include "logging/logFileOutput.hpp"
 #include "logging/logHandle.hpp"
 #include "runtime/atomic.hpp"
+#include "runtime/os.inline.hpp"
 
 class AsyncLogWriter::AsyncLogLocker : public StackObj {
  public:
@@ -51,7 +52,8 @@ void AsyncLogWriter::enqueue_locked(const AsyncLogMessage& msg) {
   }
 
   _buffer.push_back(msg);
-  _sem.signal(1, true);
+  _data_available = true;
+  _cv.notify();
 }
 
 void AsyncLogWriter::enqueue(LogFileOutput& output, const LogDecorations& decorations, const char* msg) {
@@ -75,7 +77,7 @@ void AsyncLogWriter::enqueue(LogFileOutput& output, LogMessageBuffer::Iterator m
 }
 
 AsyncLogWriter::AsyncLogWriter()
-  : _lock(1), _sem(0), _flush_sem(0),
+  : _lock(1), _flush_sem(0), _cv(), _data_available(false),
     _initialized(false),
     _stats(17 /*table_size*/) {
   if (os::create_thread(this, os::asynclog_thread)) {
@@ -125,6 +127,7 @@ void AsyncLogWriter::write() {
     // append meta-messages of dropped counters
     AsyncLogMapIterator dropped_counters_iter(logs);
     _stats.iterate(&dropped_counters_iter);
+    _data_available = false;
   }
 
   LinkedListIterator<AsyncLogMessage> it(logs.head());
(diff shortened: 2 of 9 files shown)


src/hotspot/share/logging/logAsyncWriter.cpp
```diff
@@ -148,14 +150,19 @@ void AsyncLogWriter::write() {
...
 
 void AsyncLogWriter::run() {
   while (true) {
     // The value of a semphore cannot be negative. Therefore, the current thread falls asleep
     // when its value is zero. It will be waken up when new messages are enqueued.
     _sem.wait();
-    write();
+
+    int n = write();
+    for (int i = n - 1; i > 0; --i) {
```
dholmes-ora, line 163: `while (--n > 0) { // pre-decrement as we already performed the first wait`

dholmes-ora (approved):
Hi Xin,

I wasn't expecting to see an attempt to balance the semaphore count on each write, but this seems to be okay.

One simplification suggested below.

Thanks,
David

dholmes-ora (changes requested):
Sorry I have to withdraw my initial review approval.

This certainly helps allow the writer thread to catch up sooner, but there is still no guarantee that it will keep up with the loggers sufficient to avoid the overflow.

What we have to ensure in addition is that the number of outstanding log messages permitted by the buffer size is always less than the max value allow by the semaphore. Unfortunately we don't know what that value is.

David

navyxliu:
> What we have to ensure in addition is that the number of outstanding log messages permitted by the buffer size is always less than the max value allow by the semaphore. Unfortunately we don't know what that value is.

On Linux, SEM_VALUE_MAX is defined in <semaphore.h>.  we can add an assertion that `assert(_buffer_max_size < SEM_VALUE_MAX) `. Currently,  it seems that hotspot needs to support 3 semaphores.
1. Posix semaphore of glibc. -- SEM_VALUE_MAX
2. Windows API  -- LONG_MAX
3. MacOS  -- I am not sure what this.

Yes, you just unveiled another issue.  I found _buffer_max_size is 91k if I use `-XX:AsyncLogBufferSize=50M`. That value is bigger than SEM_VALUE_MAX of AIX.

`The maximum entries of AsyncLogBuffer: 91022`

dholmes-ora:
I don't like the idea of trying to limit the log size based on a platform specific private implementation detail - and you'd have to check that against any user supplied buffer size and reject a too large buffer.

The other option is revert these changes and expose a way to ignore the ERANGE error in Semaphore.

src/hotspot/os/windows/semaphore_windows.cpp
```diff
@@ -39,7 +39,7 @@ WindowsSemaphore::~WindowsSemaphore() {
   ::CloseHandle(_semaphore);
 }
 
-void WindowsSemaphore::signal(uint count) {
+void WindowsSemaphore::signal(uint count, bool ignore_overflow) {
```
dholmes-ora, line 42: I believe the Windows ReleaseSemaphore can also fail with the equivalent of EOVERFLOW if the count would exceed the maximum allowed. We either need to dig deeper into what value of GetlastError we would see, or else ignore all errors if ignore_overflow is set. I think I can tweak a gtest to see what happens.

src/hotspot/share/logging/logAsyncWriter.hpp
```diff
@@ -107,6 +107,23 @@ class AsyncLogMessage {
   char* message() const { return _message; }
 };
 
+// The only difference between this and Semaphore is that
+// it ignores errno ERANGE and EOVERFLOW.
+class AsyncLogSemaphore : public CHeapObj<mtSynchronizer> {
+  SemaphoreImpl _impl;
```
dholmes-ora, line 113: I don't think this is necessary, just call wait(1, true) where needed.

dholmes-ora (changes requested):
I've been thinking about how to handle the ERANGE/EOVERFLOW in the Semaphore API. A default arg is one way. I was also considering a specific signal_xx method that allows overflow; or report to the caller that overflow happened. Not sure what might be best. What you have might be simplest for 17 but there are still some issues that need further investigation. See comments below. I think we need other opinions here too.

I'm surprised macOS has no overflow checks, it looks like its counter will just wrap when it overflows - which would break things, but of course is very unlikely in real use.

Thanks,
David

dholmes-ora:
I ran a test on Linux, Windows and OSX. As expected Linux gives EOVERFLOW. OSX the test just times out without error.

On Windows we get error 298:

ERROR_TOO_MANY_POSTS
    298 (0x12A)
    Too many posts were made to a semaphore.

so I think we should check for that and ignore it if ignoring overflow.

Thanks,
David

navyxliu:
got it. I will cover Windows.

navyxliu:
> I'm surprised macOS has no overflow checks, it looks like its counter will just wrap when it overflows - which would break things, but of course is very unlikely in real use.

Unlike Posix and Win32, it's not easy for me to find any document about Darwin semaphore API. 

I guess this is the equivalence of glibc on MacOS.  check out `semaphore_signal_internal`.
https://opensource.apple.com/source/xnu/xnu-344.21.74/osfmk/kern/sync_sema.c.auto.html

if count has overflown, it should return kr from `wait_queue_wakeup64_thread_locked`.

src/hotspot/share/runtime/semaphore.hpp
```diff
@@ -50,7 +50,12 @@ class Semaphore : public CHeapObj<mtSynchronizer> {
   Semaphore(uint value = 0) : _impl(value) {}
   ~Semaphore() {}
 
-  void signal(uint count = 1) { _impl.signal(count); }
+  void signal(uint count = 1) { _impl.signal(count, false /* ignore_overflow */); }
```
dholmes-ora, line 53: Please change comment to:

`/* don't ignore overflow */`

as that describes what the `false` argument actually means, rather than just showing what parameter it is providing the value for.

src/hotspot/share/runtime/semaphore.hpp
```diff
@@ -50,7 +50,12 @@ class Semaphore : public CHeapObj<mtSynchronizer> {
...
+
+  // Ignore error of overflow
+  void signal_overflow(uint count = 1) {
```
dholmes-ora, line 56: I'm a bit unsure about how to name and describe this given that the overflow check only relates to debug builds - in a product build both `signal` and `signal_overflow` are identical.

I had missed previously that the `ignore_overflow` parameter was only in the impl API not the main public API.

dholmes-ora (changes requested):
Figuring out the best way to handle this is proving to be quite tricky - sorry. I need to think some more about it. Unfortunately I'm away for a few days after today.

David

src/hotspot/share/runtime/semaphore.hpp
navyxliu: Thanks you for reviewing this change. 

The reason we only had ignore_overflow in impl API because I defined my own platform-independent Semaphore class, aka. AsyncLogSemaphore in previous revision.  I changed type of _sem from Sempahore to AsyncLogSemaphore. I "overloaded" signal(uint count) in that way so  I don't  pollute "public API".


> I don't think this is necessary, just call wait(1, true) where needed.

I think your opinion is that it's not necessary.  how about I just add a second parameter ignore_overflow to Semaphore::signal with the default value = false?

navyxliu:
Hi, @tstuefe , 
Do you mind trying this patch? David and I think we should ignore `sem_post` error when its value has been overflown. It happens rarely and ignoring this error won't affect correctness.    I just wonder if this approach can  solve the LogConfigurationTest.reconfigure_decorators_MT_vm  failure on AIX? 

> Figuring out the best way to handle this is proving to be quite tricky 

If it works out, the one open issue is what  we should do on MacOS. My option is that we just leave it alone since `semaphore_signal`(not posix semaphore) on Darwin doesn't overflow.  We can revisit this later if the assumption turns out false in the future.

tstuefe:
Sorry for dropping out, I'm snowed in. I put the patch into our AIX CI. Due to the slowness of our AIX hardware, this may take a day or two.

navyxliu:
Thanks! Let's see the result. Meanwhile, allow me refactor this patch and spend more time semaphore of [Darwin-XNU](https://github.com/apple/darwin-xnu)

navyxliu:
> One possibility is to change the code to use a Monitor that acts as the
asynclog lock and the wakeup mechanism, with a suitable guard in the
writing code so that we don't try to use the Monitor until after it has
been initialised (which is after the first log writes may occur).

I am not sure about that. The current code derived from there. HotSpot Monitor needs Thread::current(), but the logging event spans from the very beginning to the very end of JVM. We have to circumvent those cases. 

On the other hand, I carefully read the implementation of semaphore_signal on Darwin. I am pretty sure it returns KERN_SUCCESS when overflow happens. I added a line of comment in the last revision. Could you take a look at that?

navyxliu:
> But we already circumvent that for async logging. We can't use async
logging until after we have called AsyncLogWriter::initialize(). So the
synchronization objects used by AsyncLogWriter can be plain
Mutex/Monitor, they don't need to be Semaphore.

That is not true. I deleted finalize() because it was complex and not necessary after we switch from Monitor to Semaphore. 
A thread may emit logs even it has deletes itself. 

I reviewed this line. https://github.com/apple/darwin-xnu/blob/main/osfmk/kern/sync_sema.c#L400
In my understanding, `semaphore_signal` will try to wait up one in the wait queue of the semaphore. if kr != KERN_SUCCESS, it will reset counter to 0 because no one is waiting on that. In our case, we should take this route.  I think the behavior is defined. 

thanks,
--lx

navyxliu:
Aha, os::PlatformMonitor is just the cross-platform condition variable, no strings attached!
I think it is a better solution. TBH,  I don't like 'ignore_overflow'. it's a hack solution. let's save it as a back-up solution. 

We still have 2 week before phase-2 rampdown.  Let me also take a look how JFR handle this case. It is said JFR also has an asynchronous buffer too.

src/hotspot/share/logging/logAsyncWriter.hpp
```diff
@@ -137,11 +137,9 @@ class AsyncLogWriter : public NonJavaThread {
   static AsyncLogWriter* _instance;
   // _lock(1) denotes a critional region.
   Semaphore _lock;
-  // _sem is a semaphore whose value denotes how many messages have been enqueued.
-  // It decreases in AsyncLogWriter::run()
-  Semaphore _sem;
   Semaphore _flush_sem;
-
+  os::PlatformMonitor _cv;
```
dholmes-ora, line 141: No this isn't what you need. The _lock is the PlatformMonitor. You lock/unlock in AsyncLogLocker for the critical regions, and then use the wait/notify operations for the coordination.

src/hotspot/share/logging/logAsyncWriter.cpp
```diff
@@ -152,9 +155,16 @@ void AsyncLogWriter::write() {
...
   while (true) {
-    // The value of a semphore cannot be negative. Therefore, the current thread falls asleep
-    // when its value is zero. It will be waken up when new messages are enqueued.
-    _sem.wait();
+    {
+      AsyncLogLocker locker;
+
+      while (!_data_available) {
+        _lock.signal();
+        _cv.wait(0/* no timeout */);
```
dholmes-ora, line 163: You can't call wait() on an unlocked monitor  - that's not how monitors are used.

dholmes-ora (changes requested):
Sorry this isn't right at all.

David

src/hotspot/share/logging/logAsyncWriter.hpp
navyxliu: you mean I should use os::PlatformMonitor _lock as a combination of mutux and cv, right?

navyxliu:
hi, @dholmes-ora @pchilano, 

Thank you for letting me know os::PlatformMonitor. I used to think Monitor is a higher-level synchronization construct than Semaphore. it turns out that os::PlatformMonitor is as low-level as semaphore. Lesson learned. 

Could you take a look at this change and I see if I use it correctly?

src/hotspot/share/logging/logAsyncWriter.hpp
```diff
@@ -135,13 +135,9 @@ class AsyncLogWriter : public NonJavaThread {
...
 
   static AsyncLogWriter* _instance;
-  // _lock(1) denotes a critional region.
-  Semaphore _lock;
-  // _sem is a semaphore whose value denotes how many messages have been enqueued.
-  // It decreases in AsyncLogWriter::run()
-  Semaphore _sem;
   Semaphore _flush_sem;
-
+  os::PlatformMonitor _lock;
```
dholmes-ora, line 139: Please add a comment before this:

// Can't use a Monitor here as we need a low-level API that can be used
// late in a thread's termination process.

dholmes-ora (approved):
Hi Xin,

This looks good. I'll run it through our testing.

I'll also try to get @pchilano to review it.

Thanks,
David

pchilano (approved):
Hi Xin,

Looks good to me too.
The only small suggestion is that you could do the pop_all() and _stats.iterate when coming out of wait() since you are already holding the monitor, and then pass the logs to write(). Either way is fine though.

Thanks,
Patricio

navyxliu:
>  The only small suggestion is that you could do the pop_all() and _stats.iterate when coming out of wait() since you are already holding the monitor, and then pass the logs to write(). Either way is fine though.

Thanks for reviewing.  I see what you mean. I take a note on that. yes, we may gain something if we merge two critical areas. 
I will take it as a follow-up.

src/hotspot/share/logging/logAsyncWriter.hpp
navyxliu: done.

navyxliu:
Hi, Reviewers,

Thanks a lot. I append a comment to explain why we choose PlatformMonitor over Semaphore.  Meanwhile, I defer performance-wise change.  Please let me know the results of test. I hope we can make it to jdk17 **safely** before the window is close.

dholmes-ora (approved):

PR #219, opened 2021-07-06
8269897: Shenandoah: Treat UNKNOWN refs access as strong

rkennke:
We've observed test failures in jcstress, see:
https://bugs.openjdk.java.net/browse/JDK-8269897

We used to treat UNKNOWN reference accesses like weak accesses. UNKNOWN is used for Unsafe, reflection and JNI accesses, where it cannot be determined at compilation-time if we are accessing a regular field or a Reflection.referent field. The rationale for treating UNKNOWN as weak was that if the reference is a regular reference, then the value would be strongly reachable anyway, and only if it is a referent field would reachability matter. However, it turns out that this assumption is wrong: the test shows that a reference that is only weakly reachable can be legitimately written into a field, thus resurrecting the reference, and when that weakly reachable reference is loaded, it would be (wrongly) filtered as NULL.

A fix is to treat UNKNOWN accesses as strong. Accessing Reference.referent via reflection, JNI or Unsafe is Bad Idea anyway.
This test shows the problem with CAS, but I believe it affects all accesses via reflection, JNI, etc.

Testing:
 - [x] the provided jcstress test
 - [x] hotspot_gc_shenandoah
 - [x] tier1

---------
### Progress

### Issue
 * [JDK-8269897](https://bugs.openjdk.java.net/browse/JDK-8269897): Shenandoah: Treat UNKNOWN refs access as strong

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/219/head:pull/219` \
`$ git checkout pull/219`

`$ git checkout pull/219` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/219/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 219`

`$ git pr show -t 219`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/219.diff">https://git.openjdk.java.net/jdk17/pull/219.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/c2/shenandoahBarrierSetC2.cpp
+++ b/src/hotspot/share/gc/shenandoah/c2/shenandoahBarrierSetC2.cpp
@@ -640,7 +640,7 @@ Node* ShenandoahBarrierSetC2::atomic_cmpxchg_val_at_resolved(C2AtomicParseAccess
       load_store = kit->gvn().transform(new DecodeNNode(load_store, load_store->get_ptr_type()));
     }
 #endif
-    load_store = kit->gvn().transform(new ShenandoahLoadReferenceBarrierNode(NULL, load_store, access.decorators()));
+    load_store = kit->gvn().transform(new ShenandoahLoadReferenceBarrierNode(NULL, load_store, access.decorators() & ~ON_UNKNOWN_OOP_REF));
     return load_store;
   }
   return BarrierSetC2::atomic_cmpxchg_val_at_resolved(access, expected_val, new_val, value_type);
--- a/src/hotspot/share/gc/shenandoah/shenandoahBarrierSet.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahBarrierSet.hpp
@@ -54,11 +54,11 @@ public:
   static bool need_keep_alive_barrier(DecoratorSet decorators, BasicType type);
 
   static bool is_strong_access(DecoratorSet decorators) {
-    return (decorators & (ON_WEAK_OOP_REF | ON_PHANTOM_OOP_REF)) == 0;
+    return (decorators & (ON_WEAK_OOP_REF | ON_PHANTOM_OOP_REF | ON_UNKNOWN_OOP_REF)) == 0;
   }
 
   static bool is_weak_access(DecoratorSet decorators) {
-    return (decorators & ON_WEAK_OOP_REF) != 0;
+    return (decorators & (ON_WEAK_OOP_REF | ON_UNKNOWN_OOP_REF)) != 0;
   }
 
   static bool is_phantom_access(DecoratorSet decorators) {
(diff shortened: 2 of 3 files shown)


shipilev (commented):
It looks okay to me, and it fixes the jcstress failures. But I see one of the `tier1` tests failing:

```
$ make run-test TEST=gc/TestReferenceClearDuringReferenceProcessing.java TEST_VM_OPTS="-XX:+UseShenandoahGC"
...
TEST RESULT: Failed. Execution failed: `main' threw exception: java.lang.RuntimeException: ref not enqueued
```
I wonder if that failure indicates that fix is still incomplete or too conservative.

rkennke:
> It looks okay to me, and it fixes the jcstress failures. But I see one of the `tier1` tests failing:
> 
> ```
> $ make run-test TEST=gc/TestReferenceClearDuringReferenceProcessing.java TEST_VM_OPTS="-XX:+UseShenandoahGC"
> ...
> TEST RESULT: Failed. Execution failed: `main' threw exception: java.lang.RuntimeException: ref not enqueued
> ```
> 
> I wonder if that failure indicates that fix is still incomplete or too conservative.

Changing behaviour for oop_load_in_heap() seems to break it (although it is not quite obvious to me how - the test does not seem to access Reference.referent itself, especially not via reflection/unsafe/etc. Maybe something internal in Reference queue does, and expects it to return NULL on unreachable.)
Let's limit the change to treat UNKNOWN as STRONG only for CAE.

shipilev:
OK, this changes for `oop_atomic_cmpxchg_not_in_heap`, but what about `oop_atomic_cmpxchg_in_heap`?

shipilev (changes requested):
Need synopsis change after limiting a fix?
Need to handle `xchg` case too. 

This patch passes `tier1` with Shenandoah for me, but fails jcstress bundle when run with C1. A relevant `ShenandoahBarrierSetC1` change is quite probably missing.

Also, probably `Unsafe` `get` and `set` require the same handling: under concurrent weak roots, it is safer to return overly strong value, not NULL for strong refs. Let it be users' problem accessing weak values with Unsafe?

src/hotspot/share/gc/shenandoah/c2/shenandoahBarrierSetC2.cpp
```diff
@@ -640,7 +640,7 @@ Node* ShenandoahBarrierSetC2::atomic_cmpxchg_val_at_resolved(C2AtomicParseAccess
       load_store = kit->gvn().transform(new DecodeNNode(load_store, load_store->get_ptr_type()));
     }
 #endif
-    load_store = kit->gvn().transform(new ShenandoahLoadReferenceBarrierNode(NULL, load_store, access.decorators()));
+    load_store = kit->gvn().transform(new ShenandoahLoadReferenceBarrierNode(NULL, load_store, access.decorators() & ~ON_UNKNOWN_OOP_REF));
```
zhengyu123, line 643: This load should go through slow path when GC is in progress, right? Should let slow path to resolve the strength? sounds there is possibility to resurrect dead oop.

rkennke:
Fixing this in jdk18 first, should backport to jdk17 later, after some testing in 18. See: https://github.com/openjdk/jdk/pull/4697

PR #222, opened 2021-07-07
8269746: C2: assert(!in->is_CFG()) failed: CFG Node with no controlling input?

rwestrel:
The backedge and the exit test of an inner strip mined loop are
removed. A safepoint is right above the inner loop exit test when that
happens. The outer strip mined loop safepoint is then processed by
igvn which finds the other safepoint. This causes the outer loop
safepoint to be replaced by the safepoint that was in the inner
loop. Then the outer loop back edge is removed by igvn. This is done
by replacing the loop exit by the control of the outer loop
safepoint. Given the outer loop safepoint was replaced by a dominating
safepoint, that logic replaces the oiter loop exit with the dominating
safepoint control input. That must be harmless most of the time, but
in this particular case a Store was between the 2 safepoints. That
Store is still reachable in the graph but its control is not.

Note: test case not included because the crash no longer reproduces (supposedly because order of node processing by igvn matters)

---------
### Progress

### Issue
 * [JDK-8269746](https://bugs.openjdk.java.net/browse/JDK-8269746): C2: assert(!in->is_CFG()) failed: CFG Node with no controlling input?

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/222/head:pull/222` \
`$ git checkout pull/222`

`$ git checkout pull/222` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/222/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 222`

`$ git pr show -t 222`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/222.diff">https://git.openjdk.java.net/jdk17/pull/222.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/callnode.cpp
+++ b/src/hotspot/share/opto/callnode.cpp
@@ -1411,8 +1411,14 @@ Node *SafePointNode::Ideal(PhaseGVN *phase, bool can_reshape) {
 Node* SafePointNode::Identity(PhaseGVN* phase) {
 
   // If you have back to back safepoints, remove one
-  if( in(TypeFunc::Control)->is_SafePoint() )
-    return in(TypeFunc::Control);
+  if (in(TypeFunc::Control)->is_SafePoint()) {
+    Node* out_c = unique_ctrl_out();
+    // This can be the safepoint of an outer strip mined loop if the inner loop's backedge was removed. Replacing the
+    // outer loop's safepoint could confuse removal of the outer loop.
+    if (out_c != NULL && !out_c->is_OuterStripMinedLoopEnd()) {
+      return in(TypeFunc::Control);
+    }
+  }
 
   // Transforming long counted loops requires a safepoint node. Do not
   // eliminate a safepoint until loop opts are over.


chhagedorn (approved):
Looks good!

Have you tried playing around with `-XX:+StressIGVN` to trigger it again?

rwestrel:
> Have you tried playing around with `-XX:+StressIGVN` to trigger it again?

Thanks for the suggestion. I tried running the test case 1000 times in a loop with -XX:+StressIGVN but the crash doesn't reproduce.

TobiHartmann (approved):
Looks reasonable to me.

chhagedorn:
> > Have you tried playing around with `-XX:+StressIGVN` to trigger it again?
> 
> Thanks for the suggestion. I tried running the test case 1000 times in a loop with -XX:+StressIGVN but the crash doesn't reproduce.

Okay, thanks for trying it!

rwestrel:
@TobiHartmann @chhagedorn thanks for the reviews

PR #225, opened 2021-07-07
8269828: corrections in some instruction patterns for KNL x86 platform

jatin-bhateja:
Currently a temporary operand with register class regD is used in rep_stos_evex and rep_stos_large_evex instruction patterns,  whose encoding blocks calls xmm_clear_mem routine which has a VEX encoded instruction in the control path.
Thus regD class should be changed to legRegD on AVX512 platforms so that operand is always allocated from lower register bank (0-15).

There is one more issue related to usage of regD operand which is a scalar operand in a vector instruction, usually such cases may result into problems during spilling/restoration since only a scalar value is dumped and restored back, but in current context its a temporary operand may never get spilled.

---------
### Progress

### Issue
 * [JDK-8269828](https://bugs.openjdk.java.net/browse/JDK-8269828): corrections in some instruction patterns for KNL x86 platform

### Reviewers
 * [Dean Long](https://openjdk.java.net/census#dlong) (@dean-long - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/225/head:pull/225` \
`$ git checkout pull/225`

`$ git checkout pull/225` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/225/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 225`

`$ git pr show -t 225`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/225.diff">https://git.openjdk.java.net/jdk17/pull/225.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/x86_32.ad
+++ b/src/hotspot/cpu/x86/x86_32.ad
@@ -11529,7 +11529,7 @@ instruct rep_stos(eCXRegI cnt, eDIRegP base, regD tmp, eAXRegI zero, Universe du
 %}
 
 // Small ClearArray AVX512 non-constant length.
-instruct rep_stos_evex(eCXRegI cnt, eDIRegP base, regD tmp, kReg ktmp, eAXRegI zero, Universe dummy, eFlagsReg cr) %{
+instruct rep_stos_evex(eCXRegI cnt, eDIRegP base, legRegD tmp, kReg ktmp, eAXRegI zero, Universe dummy, eFlagsReg cr) %{
   predicate(!((ClearArrayNode*)n)->is_large() && (UseAVX > 2));
   match(Set dummy (ClearArray cnt base));
   ins_cost(125);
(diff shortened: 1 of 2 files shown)


vnkozlov (commented):
Seems reasonable. Let me test it.

dean-long:
@jatin-bhateja Could you explain more about the spilling/restoration problem?  I guess this fix solves both problems, at least for ClearArray, but what about other instructions?

dean-long (approved):

vnkozlov (approved):
hs-tier1-3 testing passed.

jatin-bhateja:
> @jatin-bhateja Could you explain more about the spilling/restoration problem? I guess this fix solves both problems, at least for ClearArray, but what about other instructions?

Hi @dean-long , each bit of register mask corresponds to 32 bit word, in case of regD operand (scalar) mask will have two bits where as for vecX operand (vector) mask will have 4 bit set. This will result into difference in spill sizes even if both these operands are allocated same physical register set (xmm0-15/31) . As per usual process separate RFE will be filed on encounteting more erroneous patterns.

PR #231, opened 2021-07-08
8261147: C2: Node is wrongly marked as reduction resulting in a wrong execution due to wrong vector instructions

chhagedorn:
C2 produces a wrong result for the testcase due to treating a non-reduction `AddL` node wrongly as a reduction (`Node::is_reduction()` is true). This node is later part of a pack in the superword algorithm which wrongly treats it as a reduction. As a result, we create an `AddReductionVL` instead of an `AddVL` node, letting the test fail.

The wrong reduction can be traced back to `AddNode::Ideal` where we create a clone of an `AddL` node which is marked as a reduction. However, the clone itself is no longer a reduction but is still marked as such. `Node::clone()` simply copies the `Flag_is_reduction` flag in `Node::_flags`:
https://github.com/openjdk/jdk17/blob/4f707591754e5e7f747d1d0a47f78f49060771c2/src/hotspot/share/opto/addnode.cpp#L181-L196

I don't think we need to clone the `Flag_is_reduction` flag by default in `Node::clone()`. I went through the uses of `Node::clone()` and I think we only need to clone the flag when copying an entire loop body in `PhaseIdealLoop::clone_loop()`.  That's what I propose as a fix. This approach also avoids the need to think about reductions in future uses of `Node::clone()` with a special handling for it as it would be required in `AddNode::Ideal` etc.

I did some performance testing with some common benchmarks which looked good. I also ran the follwing JTreg tests which showed the same number of newly created vectors with and without the fix:
```
jtreg -testjdk:jdk-18/fastdebug -va -javaoptions:'-server -Xbatch -XX:-TieredCompilation -XX:CICompilerCount=1 -XX:+TraceNewVectors' -J-Djavatest.maxOutputSize=1000000 compiler/c2/cr6340864/ compiler/codegen/ compiler/loopopts/superword/ compiler/vectorization > new_vects.log
grep "new Vector node:" new_vects.log | wc
  20190  521206 4398994
```

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8261147](https://bugs.openjdk.java.net/browse/JDK-8261147): C2: Node is wrongly marked as reduction resulting in a wrong execution due to wrong vector instructions

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/231/head:pull/231` \
`$ git checkout pull/231`

`$ git checkout pull/231` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/231/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 231`

`$ git pr show -t 231`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/231.diff">https://git.openjdk.java.net/jdk17/pull/231.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/loopopts.cpp
+++ b/src/hotspot/share/opto/loopopts.cpp
@@ -2203,10 +2203,14 @@ void PhaseIdealLoop::clone_loop( IdealLoopTree *loop, Node_List &old_new, int dd
 
   // Step 1: Clone the loop body.  Make the old->new mapping.
   uint i;
-  for( i = 0; i < loop->_body.size(); i++ ) {
-    Node *old = loop->_body.at(i);
-    Node *nnn = old->clone();
-    old_new.map( old->_idx, nnn );
+  for (i = 0; i < loop->_body.size(); i++) {
+    Node* old = loop->_body.at(i);
+    Node* nnn = old->clone();
+    old_new.map(old->_idx, nnn);
+    if (old->is_reduction()) {
+      // Reduction flag is not copied by default. Copy it here when cloning the entire loop body.
+      nnn->add_flag(Node::Flag_is_reduction);
+    }
     if (C->do_vector_loop()) {
       cm.verify_insert_and_clone(old, nnn, cm.clone_idx());
     }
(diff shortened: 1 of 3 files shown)


TobiHartmann (approved):
Looks good to me.

chhagedorn:
Thanks Tobias for your review!

vnkozlov (approved):
Good.

chhagedorn:
Thanks Vladimir for your review!

PR #235, opened 2021-07-08
8269795: C2: Out of bounds array load floats above its range check in loop peeling resulting in SEGV

chhagedorn:
In the testcase, an out of bounds array load (`LoadI`) floats above its range check resulting in a segfault. The problem can be traced back to incorrectly treating an `If` node, on which the `LoadI` is control dependent, as being dominated in `PhaseIdealLoop::peeled_dom_test_elim`. As a result, `PhaseIdealLoop::dominated_by()` moves the `LoadI` out and before the loop while the range check stays inside the loop.

`peeled_dom_test_elim()` walks through the idom chain of the loop body and tries to find loop-invariant tests. Once it finds one, it not only moves this test out of the loop but also looks for other tests with the same condition node which can be removed. The condition of the dominated tests in the loop are replaced with a `ConI` in `dominated_by()`. The data dependencies on the removed dominated tests are rewired to a peeled node before the loop.

However, the current code directly checks `test->in(1)` at L527 which is only the initial `Bool` node of the dominated test before the first invocation of `dominated_by()` in the loop on L525. Afterwards, `test->in(1)` is a `ConI` node. We now continue to look for tests with an identical `ConI` condition to move their data nodes out of this loop. This is wrong and exactly happens in the testcase: The loop body still contains a not yet cleaned up (by IGVN) `If` node (with a `ConI` condition) of an eliminated skeleton predicate on which an out of bounds  `LoadI` node is control dependent. The `LoadI` is then wrongly rewired out of the loop and before its range check.

The fix is to only check against the initial `test->in(1)` `Bool` node which satisfies the conditions checked on the lines before (L518-520).

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8269795](https://bugs.openjdk.java.net/browse/JDK-8269795): C2: Out of bounds array load floats above its range check in loop peeling resulting in SEGV

### Reviewers
 * [Roland Westrelin](https://openjdk.java.net/census#roland) (@rwestrel - **Reviewer**) ⚠️ Review applies to c29fd2470626506e8801414e028230665d502c85

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/235/head:pull/235` \
`$ git checkout pull/235`

`$ git checkout pull/235` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/235/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 235`

`$ git pr show -t 235`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/235.diff">https://git.openjdk.java.net/jdk17/pull/235.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/loopTransform.cpp
+++ b/src/hotspot/share/opto/loopTransform.cpp
@@ -507,24 +507,24 @@ uint IdealLoopTree::estimate_peeling(PhaseIdealLoop *phase) {
 // If we got the effect of peeling, either by actually peeling or by making
 // a pre-loop which must execute at least once, we can remove all
 // loop-invariant dominated tests in the main body.
-void PhaseIdealLoop::peeled_dom_test_elim(IdealLoopTree *loop, Node_List &old_new) {
+void PhaseIdealLoop::peeled_dom_test_elim(IdealLoopTree* loop, Node_List& old_new) {
   bool progress = true;
   while (progress) {
-    progress = false;           // Reset for next iteration
-    Node *prev = loop->_head->in(LoopNode::LoopBackControl);//loop->tail();
-    Node *test = prev->in(0);
+    progress = false; // Reset for next iteration
+    Node* prev = loop->_head->in(LoopNode::LoopBackControl); // loop->tail();
+    Node* test = prev->in(0);
     while (test != loop->_head) { // Scan till run off top of loop
-
       int p_op = prev->Opcode();
       if ((p_op == Op_IfFalse || p_op == Op_IfTrue) &&
-          test->is_If() &&      // Test?
+          test->is_If() && // Test?
           !test->in(1)->is_Con() && // And not already obvious?
           // Condition is not a member of this loop?
           !loop->is_member(get_loop(get_ctrl(test->in(1))))){
         // Walk loop body looking for instances of this test
+        Node* test_cond = test->in(1);
         for (uint i = 0; i < loop->_body.size(); i++) {
-          Node *n = loop->_body.at(i);
-          if (n->is_If() && n->in(1) == test->in(1) /*&& n != loop->tail()->in(0)*/) {
+          Node* n = loop->_body.at(i);
+          if (n->is_If() && n->in(1) == test_cond) {
             // IfNode was dominated by version in peeled loop body
             progress = true;
             dominated_by(old_new[prev->_idx], n);
@@ -534,7 +534,6 @@ void PhaseIdealLoop::peeled_dom_test_elim(IdealLoopTree *loop, Node_List &old_ne
       prev = test;
       test = idom(test);
     } // End of scan tests in loop
-
   } // End of while (progress)
 }
 
--- a/src/hotspot/share/opto/loopTransform.cpp
+++ b/src/hotspot/share/opto/loopTransform.cpp
@@ -515,15 +515,20 @@ void PhaseIdealLoop::peeled_dom_test_elim(IdealLoopTree* loop, Node_List& old_ne
     Node* test = prev->in(0);
     while (test != loop->_head) { // Scan till run off top of loop
       int p_op = prev->Opcode();
-      if ((p_op == Op_IfFalse || p_op == Op_IfTrue) &&
-          test->is_If() && // Test?
-          !test->in(1)->is_Con() && // And not already obvious?
-          // Condition is not a member of this loop?
-          !loop->is_member(get_loop(get_ctrl(test->in(1))))){
+      assert(test != NULL, "test cannot be NULL");
+      Node* test_cond = NULL;
+      if ((p_op == Op_IfFalse || p_op == Op_IfTrue) && test->is_If()) {
+        test_cond = test->in(1);
+      }
+      if (test_cond != NULL && // Test?
+          !test_cond->is_Con() && // And not already obvious?
+          // And condition is not a member of this loop?
+          !loop->is_member(get_loop(get_ctrl(test_cond)))) {
         // Walk loop body looking for instances of this test
-        Node* test_cond = test->in(1);
         for (uint i = 0; i < loop->_body.size(); i++) {
           Node* n = loop->_body.at(i);
+          // Check against cached test condition because dominated_by()
+          // replaces the test condition with a constant.
           if (n->is_If() && n->in(1) == test_cond) {
             // IfNode was dominated by version in peeled loop body
             progress = true;
--- a/test/hotspot/jtreg/compiler/loopopts/TestPeelingRemoveDominatedTest.java
+++ b/test/hotspot/jtreg/compiler/loopopts/TestPeelingRemoveDominatedTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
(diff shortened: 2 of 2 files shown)


test/hotspot/jtreg/compiler/loopopts/TestPeelingRemoveDominatedTest.java
```diff
@@ -0,0 +1,67 @@
...
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @key stress randomness
```
TobiHartmann, line 26: There is no randomness in the test, right? Or is that referring to the use of `StressGCM`? If so, these keys are missing from other other tests that use this flag.

test/hotspot/jtreg/compiler/loopopts/TestPeelingRemoveDominatedTest.java
```diff
@@ -0,0 +1,67 @@
...
+    static int iArrFld[] = new int[N];
+
+    public static void main(String[] strArr) {
+        TestPeelingRemoveDominatedTest _instance = new TestPeelingRemoveDominatedTest();
+        for (int i = 0; i < 10; i++) {
+            _instance.mainTest();
+        }
+    }
+
+    public void mainTest() {
```
TobiHartmann, line 49: Is that method needed?

src/hotspot/share/opto/loopTransform.cpp
```diff
@@ -507,24 +507,24 @@ uint IdealLoopTree::estimate_peeling(PhaseIdealLoop *phase) {
...
-
       int p_op = prev->Opcode();
       if ((p_op == Op_IfFalse || p_op == Op_IfTrue) &&
-          test->is_If() &&      // Test?
+          test->is_If() && // Test?
           !test->in(1)->is_Con() && // And not already obvious?
           // Condition is not a member of this loop?
           !loop->is_member(get_loop(get_ctrl(test->in(1))))){
         // Walk loop body looking for instances of this test
+        Node* test_cond = test->in(1);
```
TobiHartmann, line 524: Maybe move this up to line 517 and replace the `test->in(1)->is_Con()` and `get_ctrl(test->in(1))` usages in line 520/522 for better readability.

TobiHartmann (approved):
Nice analysis, the fix looks good to me! Just some minor comments below.

> [...] which is only the initial Bool node of the dominating test before the first invocation [...]

Should be "of the **dominated** test", right?

src/hotspot/share/opto/loopTransform.cpp
chhagedorn: Unfortunately, that's not possible as `test` could be an `IfProj` with only a control input.

test/hotspot/jtreg/compiler/loopopts/TestPeelingRemoveDominatedTest.java
chhagedorn: Yes, it's referring to `StressGCM`. @robcasloz once updated all tests with `StressGCM` to use these keys in [JDK-8253765](https://github.com/openjdk/jdk/commit/05459df0). But some newer tests miss these keys. Should we update them in an RFE to be consistent?

test/hotspot/jtreg/compiler/loopopts/TestPeelingRemoveDominatedTest.java
chhagedorn: It is needed, otherwise, it does not reproduce.

chhagedorn:
> Nice analysis, the fix looks good to me! Just some minor comments below.

Thanks for your review Tobias!

> 
> > [...] which is only the initial Bool node of the dominating test before the first invocation [...]
> 
> Should be "of the **dominated** test", right?

Yes, you're right as we are updating `test` to nodes in the loop.

test/hotspot/jtreg/compiler/loopopts/TestPeelingRemoveDominatedTest.java
TobiHartmann: Yes, I think that would be good.

src/hotspot/share/opto/loopTransform.cpp
TobiHartmann: Right, too bad.

test/hotspot/jtreg/compiler/loopopts/TestPeelingRemoveDominatedTest.java
chhagedorn: Okay, I filed [JDK-8270156](https://bugs.openjdk.java.net/browse/JDK-8270156) to clean it up later.

rwestrel (approved):
Looks good to me.

chhagedorn:
Thanks Roland for your review!

src/hotspot/share/opto/loopTransform.cpp
vnkozlov: Add comment why you cache the value - `dominated_by() call may modify it`.

src/hotspot/share/opto/loopTransform.cpp
```diff
@@ -507,24 +507,24 @@ uint IdealLoopTree::estimate_peeling(PhaseIdealLoop *phase) {
...
-void PhaseIdealLoop::peeled_dom_test_elim(IdealLoopTree *loop, Node_List &old_new) {
+void PhaseIdealLoop::peeled_dom_test_elim(IdealLoopTree* loop, Node_List& old_new) {
   bool progress = true;
   while (progress) {
-    progress = false;           // Reset for next iteration
-    Node *prev = loop->_head->in(LoopNode::LoopBackControl);//loop->tail();
-    Node *test = prev->in(0);
+    progress = false; // Reset for next iteration
+    Node* prev = loop->_head->in(LoopNode::LoopBackControl); // loop->tail();
+    Node* test = prev->in(0);
```
vnkozlov, line 515: May be add `assert(test != NULL)`. `idom()` call at the end of loop has such asserts.
In other places we have check `test != NULL` but we not doing that here (at line 519).

src/hotspot/share/opto/loopTransform.cpp
```diff
@@ -507,24 +507,24 @@ uint IdealLoopTree::estimate_peeling(PhaseIdealLoop *phase) {
...
     while (test != loop->_head) { // Scan till run off top of loop
-
       int p_op = prev->Opcode();
       if ((p_op == Op_IfFalse || p_op == Op_IfTrue) &&
-          test->is_If() &&      // Test?
+          test->is_If() && // Test?
```
vnkozlov, line 519: We can do what Tobias is asking by pre-check:
```
  int p_op = prev->Opcode();
  Node* test_cond = NULL;
  if ((p_op == Op_IfFalse || p_op == Op_IfTrue) && test->is_If()) {
    test_cond = test->in(1);
  }
  if (test_cond != NULL &&
```

vnkozlov (commented):
Looks good. I have few comments.

vnkozlov (approved):
Good.

TobiHartmann (approved):
Looks good!

chhagedorn:
Thanks Vladimir, Tobias and Roland for your reviews!

PR #237, opened 2021-07-09
8270151: IncompatibleClassChangeError on empty pattern switch statement case

lahodaj:
Consider code like:
```
sealed interface I {}
final class A implements I {}
...
I i = null;
switch (i) {
     case A a:
}
```

The switch is exhaustive, and javac must generate a synthetic default clause throwing an `IncompatibleClassChangeError` for the case where the permitted classes for I changes. The default clause is generated at the end of the switch statement, and consequently the execution will fall through from `case A a:` to this synthetic default, causing the `IncompatibleClassChangeError` for a valid input. (Note this was not a problem for switch expressions, as the last case can never complete normally for a switch expression.)

This PR proposes to put the synthetic default at the beginning of the switch statement, rather than at the end, which should avoid fall-through issues. An alternative solution would be to generate a synthetic break into the last case, but that is a bit tricky, so placing the default at the beginning seems more reliable to me.

---------
### Progress

### Issue
 * [JDK-8270151](https://bugs.openjdk.java.net/browse/JDK-8270151): IncompatibleClassChangeError on empty pattern switch statement case

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/237/head:pull/237` \
`$ git checkout pull/237`

`$ git checkout pull/237` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/237/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 237`

`$ git pr show -t 237`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/237.diff">https://git.openjdk.java.net/jdk17/pull/237.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Lower.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Lower.java
@@ -3598,7 +3598,7 @@ public class Lower extends TreeTranslator {
             JCThrow thr = make.Throw(makeNewClass(syms.incompatibleClassChangeErrorType,
                                                   List.nil()));
             JCCase c = make.Case(JCCase.STATEMENT, List.of(make.DefaultCaseLabel()), List.of(thr), null);
-            cases = cases.append(c);
+            cases = cases.prepend(c);
         }
 
         return cases;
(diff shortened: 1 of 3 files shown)


mcimadamore (approved):
Looks good!

PR #247, opened 2021-07-13
8269752: C2: assert(false) failed: Bad graph detected in build_loop_late

rwestrel:
Some nodes are sunk out of the outer loop of vMeth(). A CastII node is
added on the exit path of the loop to pin the nodes out of the
loop. The dominating test of that CastII is:

if (i >= 132) {

so the CastII is updated with type int [132..]

The inner loop is eliminated, the outer loop is turned into a counted
loop, pre/main/post loops are created. The CastII control is now a
Region that merges paths from the pre/main/post loops. Its data input
is a Phi that merges the iv Phis of the 3 loops.

The loop body includes an early exit:

if (i2 != 0) {
    vMeth_check_sum += i3;
    return;
}

That test is guaranteed to be taken at the second iteration. As a
consequence, the backedge of the main loop is removed. What's left of
the main loop is still guarded by a an entry test that compares the iv
value out of the pre loop with an Opaque1 node. The CastII is now only
reachable through that test. The iv Phi of the pre loop constant folds
to a value that's not >= 132. As a consequence the CastII becomes
top. top propagates to other data node and as a consequence a Phi is
replaced by its only non top input. That causes the assert failure.

The CastII should not be reachable anymore once the iv Phi of the pre
loop has constant folded and the early exit in the main loop is always
taken. But because the test that guards the main loop uses an Opaque1,
it cannot constant fold. As a fix, I propose removing the Opaque1 once
the backedge of the main loop is removed. This allows the entry test
to constant fold. The Opaque1 has no purpose anymore given the loop
doesn't exist at this point.

---------
### Progress

### Issue
 * [JDK-8269752](https://bugs.openjdk.java.net/browse/JDK-8269752): C2: assert(false) failed: Bad graph detected in build_loop_late

### Reviewers
 * [Christian Hagedorn](https://openjdk.java.net/census#chagedorn) (@chhagedorn - **Reviewer**) ⚠️ Review applies to 52028ab3b1f0b0e72944fff8077daa4e0cd43e71

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/247/head:pull/247` \
`$ git checkout pull/247`

`$ git checkout pull/247` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/247/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 247`

`$ git pr show -t 247`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/247.diff">https://git.openjdk.java.net/jdk17/pull/247.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/loopnode.cpp
+++ b/src/hotspot/share/opto/loopnode.cpp
@@ -2139,9 +2139,10 @@ SafePointNode* CountedLoopNode::outer_safepoint() const {
 }
 
 Node* CountedLoopNode::skip_predicates_from_entry(Node* ctrl) {
-    while (ctrl != NULL && ctrl->is_Proj() && ctrl->in(0)->is_If() &&
-           ctrl->in(0)->as_If()->proj_out(1-ctrl->as_Proj()->_con)->outcnt() == 1 &&
-           ctrl->in(0)->as_If()->proj_out(1-ctrl->as_Proj()->_con)->unique_out()->Opcode() == Op_Halt) {
+    while (ctrl != NULL && ctrl->is_Proj() && ctrl->in(0) != NULL && ctrl->in(0)->is_If() &&
+            (ctrl->in(0)->as_If()->proj_out_or_null(1-ctrl->as_Proj()->_con) == NULL ||
+             (ctrl->in(0)->as_If()->proj_out(1-ctrl->as_Proj()->_con)->outcnt() == 1 &&
+              ctrl->in(0)->as_If()->proj_out(1-ctrl->as_Proj()->_con)->unique_out()->Opcode() == Op_Halt))) {
       ctrl = ctrl->in(0)->in(0);
     }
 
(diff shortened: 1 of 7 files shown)


chhagedorn (approved):
Nice analysis, that looks reasonable to me.

rwestrel:
@chhagedorn  thanks for the review

chhagedorn (approved):

src/hotspot/share/opto/loopnode.cpp
```diff
@@ -4920,28 +4921,32 @@ Node* PhaseIdealLoop::compute_lca_of_uses(Node* n, Node* early, bool verify) {
...
   }
   Node* cmpzm = bolzm->in(1);
   if (cmpzm == NULL || !cmpzm->is_Cmp()) {
-    return false;
+    return NULL;
   }
   // compares can get conditionally flipped
+
+  int input = is_main_loop() ? 2 : 1;
+  bool res = cmpzm->in(input)->Opcode() == Op_Opaque1;
```
vnkozlov, line 4948: You need to check `input < cmpzm->in(input)` and  `cmpzm->in(input)` for `NULL`. There could be a "dead" path.

rwestrel: Thanks for reviewing this. I updated the patch.

vnkozlov (approved):
Good.

PR #252, opened 2021-07-16
8270461: ZGC: Invalid oop passed to ZBarrierSetRuntime::load_barrier_on_oop_array

TobiHartmann:
For object arrays, C2's clone intrinsic emits calls to the `oop_disjoint_arraycopy_uninit` stub. With ZGC, load barriers on the source array elements are applied via `BarrierSetAssembler::arraycopy_prologue` before copying to the destination array: https://github.com/openjdk/jdk17/blob/1e3b418a53a080a53827989393362338b43dd363/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp#L2400-L2403

The problem is that `BarrierSetC2::arraycopy_payload_base_offset` may 8-byte align the array offset to ensure that a copy in 8-byte chunks of the 8-byte aligned object is possible: https://github.com/openjdk/jdk17/blob/1e3b418a53a080a53827989393362338b43dd363/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp#L658-L662
Now with ` -XX:-UseCompressedClassPointers`, the offset starts at the 4-byte length field of the array (8 bytes mark word + 8 byte klass = 16 byte). That is fine if we don't need load barriers and can copy the array as `T_LONG` but with ZGC we crash in `ZBarrier::mark` because the element is not a valid oop.

I propose to simply set the offset to the first array element when cloning Object arrays with ZGC. We can still copy in 8 byte chunks because the oop elements are 8 byte on 64-bit (and ZGC is only supported on 64-bit).

I found this when investigating intermittent ZGC crashes in project Valhalla. The bug was introduced by [JDK-8268125](https://bugs.openjdk.java.net/browse/JDK-8268125) in JDK 17. The code is quite messy and will hopefully be cleaned up by [JDK-8268020](https://bugs.openjdk.java.net/browse/JDK-8268020).

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8270461](https://bugs.openjdk.java.net/browse/JDK-8270461): ZGC: Invalid oop passed to ZBarrierSetRuntime::load_barrier_on_oop_array

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/252/head:pull/252` \
`$ git checkout pull/252`

`$ git checkout pull/252` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/252/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 252`

`$ git pr show -t 252`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/252.diff">https://git.openjdk.java.net/jdk17/pull/252.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/z/c2/zBarrierSetC2.cpp
+++ b/src/hotspot/share/gc/z/c2/zBarrierSetC2.cpp
@@ -296,6 +296,13 @@ void ZBarrierSetC2::clone_at_expansion(PhaseMacroExpand* phase, ArrayCopyNode* a
     Node* dest_offset = ac->in(ArrayCopyNode::DestPos);
     Node* length = ac->in(ArrayCopyNode::Length);
 
+    if (bt == T_OBJECT) {
+      // BarrierSetC2::arraycopy_payload_base_offset 8-byte aligns the offset.
+      // Make sure the offset points to the first element in the array when cloning
+      // object arrays. Otherwise, load barriers are applied to parts of the header.
+      src_offset = phase->longcon(arrayOopDesc::base_offset_in_bytes(T_OBJECT));
+      dest_offset = src_offset;
+    }
     Node* payload_src = phase->basic_plus_adr(src, src_offset);
     Node* payload_dst = phase->basic_plus_adr(dest, dest_offset);
 
--- a/src/hotspot/share/gc/z/c2/zBarrierSetC2.cpp
+++ b/src/hotspot/share/gc/z/c2/zBarrierSetC2.cpp
@@ -297,9 +297,10 @@ void ZBarrierSetC2::clone_at_expansion(PhaseMacroExpand* phase, ArrayCopyNode* a
     Node* length = ac->in(ArrayCopyNode::Length);
 
     if (bt == T_OBJECT) {
-      // BarrierSetC2::arraycopy_payload_base_offset 8-byte aligns the offset.
-      // Make sure the offset points to the first element in the array when cloning
-      // object arrays. Otherwise, load barriers are applied to parts of the header.
+      // BarrierSetC2::clone sets the offsets via BarrierSetC2::arraycopy_payload_base_offset
+      // which 8-byte aligns them to allow for word size copies. Make sure the offsets point
+      // to the first element in the array when cloning object arrays. Otherwise, load
+      // barriers are applied to parts of the header.
       assert(src_offset == dest_offset, "should be equal");
       assert((src_offset->get_long() == arrayOopDesc::base_offset_in_bytes(T_OBJECT) && UseCompressedClassPointers) ||
              (src_offset->get_long() == arrayOopDesc::length_offset_in_bytes() && !UseCompressedClassPointers),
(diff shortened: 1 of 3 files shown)


vnkozlov:
Should we fix `arraycopy_payload_base_offset()` instead? The placement of  the fix looks strange: `src_offset` and `dest_offset` are some input nodes from `ArrayCopyNode` but you replacing them with one constant.
The fix is valid because the code is guarded by `ac->is_clone_array()` check. But I think `src_offset` and `dest_offset` should be already corrected when `ZBarrierSetC2::clone_at_expansion()` is called.

TobiHartmann:
Thanks for the review, Vladimir!

The problem is that in `arraycopy_payload_base_offset` we don't know the basic type and therefore can't use `arrayOopDesc::base_offset_in_bytes(bt)`:
https://github.com/openjdk/jdk17/blob/1e3b418a53a080a53827989393362338b43dd363/src/hotspot/share/gc/shared/c2/barrierSetC2.cpp#L660

Also, modifying that code would affect the other GCs using `T_LONG` for cloning Object arrays. Since we are late for JDK 17, I would rather prefer going with this conservative fix and postpone cleaning things up to [JDK-8268020](https://bugs.openjdk.java.net/browse/JDK-8268020). What do you think?

And yes, for `ArrayCopyNodes` used for clone, `src_offset` is always a constant and equal to `dest_offset` (see code in `BarrierSetC2::clone`).

src/hotspot/share/gc/z/c2/zBarrierSetC2.cpp
```diff
@@ -296,6 +296,13 @@ void ZBarrierSetC2::clone_at_expansion(PhaseMacroExpand* phase, ArrayCopyNode* a
     Node* dest_offset = ac->in(ArrayCopyNode::DestPos);
     Node* length = ac->in(ArrayCopyNode::Length);
 
+    if (bt == T_OBJECT) {
+      // BarrierSetC2::arraycopy_payload_base_offset 8-byte aligns the offset.
+      // Make sure the offset points to the first element in the array when cloning
+      // object arrays. Otherwise, load barriers are applied to parts of the header.
+      src_offset = phase->longcon(arrayOopDesc::base_offset_in_bytes(T_OBJECT));
```
vnkozlov, line 303: Add comment that `BarrierSetC2::clone()` uses `arraycopy_payload_base_offset(is_array);` to set src and dst offsets. And add assert that `src_offset` and `dest_offset` are indeed equal to that constant.

My concern is that `CloneArray` could be set in other places in a future (currently only `BarrierSetC2::clone()` uses it) and offset could be different (may be already set correct values as you do).

Don't forget about RDP2 process.

TobiHartmann: Thanks Vladimir. I've adjusted the comments and added asserts.

I'll request approval for integration into JDK 17.

chhagedorn (approved):
Looks reasonable.

TobiHartmann:
Thanks, Christian!

vnkozlov (approved):
Good.

TobiHartmann:
Thanks, Vladimir!

PR #255, opened 2021-07-19
8269820: C2 PhaseIdealLoop::do_unroll get wrong opaque node

huishi-hs:
More discussion in previous PR(https://github.com/openjdk/jdk17/pull/208) is confilict with fix for JDK-8269752.

Opaque1 in main loop entry is expected only used in compare node's second input. However split if might clone opaque1 and replace original node with phi of opaque1 node. This causes assertion in CountedLoopNode::is_canonical_loop_entry now.

Fix is in BoolNode::Ideal, avoiding switching compare node's input when second input is phi of opaque1.

Test: Linux X64 tier1/2/3 release/fastdebug no regression.

---------
### Progress

### Issue
 * [JDK-8269820](https://bugs.openjdk.java.net/browse/JDK-8269820): C2 PhaseIdealLoop::do_unroll get wrong opaque node

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/255/head:pull/255` \
`$ git checkout pull/255`

`$ git checkout pull/255` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/255/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 255`

`$ git pr show -t 255`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/255.diff">https://git.openjdk.java.net/jdk17/pull/255.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/subnode.cpp
+++ b/src/hotspot/share/opto/subnode.cpp
@@ -1396,6 +1396,18 @@ static bool is_counted_loop_cmp(Node *cmp) {
          n->in(0)->as_CountedLoop()->phi() == n;
 }
 
+static bool is_phi_has_opaque1(Node* n) {
+  if (n->is_Phi()) {
+    for (uint i = 1; i < n->req(); i++) {
+      Node* in = n->in(i);
+      if (in != NULL && in->is_Opaque1()) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
 //------------------------------Ideal------------------------------------------
 Node *BoolNode::Ideal(PhaseGVN *phase, bool can_reshape) {
   // Change "bool tst (cmp con x)" into "bool ~tst (cmp x con)".
@@ -1418,7 +1430,7 @@ Node *BoolNode::Ideal(PhaseGVN *phase, bool can_reshape) {
   // Move constants to the right of compare's to canonicalize.
   // Do not muck with Opaque1 nodes, as this indicates a loop
   // guard that cannot change shape.
-  if( con->is_Con() && !cmp2->is_Con() && op2 != Op_Opaque1 &&
+  if( con->is_Con() && !cmp2->is_Con() && op2 != Op_Opaque1 && !is_phi_has_opaque1(cmp2) &&
       // Because of NaN's, CmpD and CmpF are not commutative
       cop != Op_CmpD && cop != Op_CmpF &&
       // Protect against swapping inputs to a compare when it is used by a
--- a/test/hotspot/jtreg/compiler/loopopts/TestCanonicalLoopEntryOpaqueOrder.java
+++ b/test/hotspot/jtreg/compiler/loopopts/TestCanonicalLoopEntryOpaqueOrder.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (C) 2021 THL A29 Limited, a Tencent company. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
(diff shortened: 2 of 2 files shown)


rwestrel:
The proposed fix is ok but I wonder how robust it is. Are we sure there can't be 2 (or more) split thru phi in a row and that the Opaque1 node is not hiding behind a chain of Phis?

Note that the bug occurs because the Opaque1 node has control above the predicates due to this:
https://github.com/openjdk/jdk17/blob/master/src/hotspot/share/opto/loopnode.cpp#L5356
That logic is not required for Opaque1 nodes AFAICT. So maybe a simpler fix would be to skip it for Opaque1 nodes.

huishi-hs:
@rwestrel Thanks for you review!

> The proposed fix is ok but I wonder how robust it is. Are we sure there can't be 2 (or more) split thru phi in a row and that the Opaque1 node is not hiding behind a chain of Phis?
> 

In normal case, opaque1 has only 1 input, it cannot split more than one time. In special case, opaque1 has two limit inputs, extra one is original loop limit for range check. But this step(IdealLoopTree::iteration_split) happens after split if, so there will not be cases opaque1 behind a chain of Phis in my own understanding. .

> Note that the bug occurs because the Opaque1 node has control above the predicates due to this:
> https://github.com/openjdk/jdk17/blob/master/src/hotspot/share/opto/loopnode.cpp#L5356
> That logic is not required for Opaque1 nodes AFAICT. So maybe a simpler fix would be to skip it for Opaque1 nodes.

Not understand how https://github.com/openjdk/jdk17/blob/master/src/hotspot/share/opto/loopnode.cpp#L5356 cause this bug.
Are you suggesting skipping split if wit Opaque1 in tree as discussed in ealier PR?  Skipping must be a safer  and conservertive fix, I can submit a new fix if safer fix is perfer now.

rwestrel:
> In normal case, opaque1 has only 1 input, it cannot split more than one time. In special case, opaque1 has two limit inputs, extra one is original loop limit for range check. But this step(IdealLoopTree::iteration_split) happens after split if, so there will not be cases opaque1 behind a chain of Phis in my own understanding. .

It's not that obvious to me. Couldn't split thru phi cause Opaque1 to be split several times thru several Regions?

> Not understand how https://github.com/openjdk/jdk17/blob/master/src/hotspot/share/opto/loopnode.cpp#L5356 cause this bug.
> Are you suggesting skipping split if wit Opaque1 in tree as discussed in ealier PR? Skipping must be a safer and conservertive fix, I can submit a new fix if safer fix is perfer now.

Here is the fix I suggest:
https://github.com/rwestrel/jdk17/commit/15f515a56777517493aee378616c0a0cda294a16

Opaque1 is split thru Phi because its control is set to the Region that dominates the predicates. The code I pointed to causes the control to be set that way. But I don't think that logic applies to Opaque1 nodes.

vnkozlov:
There could be a case when Opaque1 is used by predicate which follows a Region without been placed there by code Roland is pointing. Then Roland's suggested fix may not work in general case.
And I understand the concern about hidden Opaque1 behind Phis. I need to look on this case to make suggestion. In general we do split_if if we can fold values on one of paths. I think we should not do it for Cmp which reference Opaque1. But I need more details.

I will look on this case today.

vnkozlov:
The issue is more complicated and I need more time to investigate it.
And I moved bug Fix version to JDK 18 (latest JDK).
This is NOT new regression in JDK 17 - it failed in previous releases.
Ffailure changed after 8269752 but on the same code path.

Please, keep current PR until we get the fix.

huishi-hs:
@vnkozlov @rwestrel Thanks for your help!

kevinrushforth:
The jdk17 repo is no longer open for any fixes. If this PR is to move forward, it will need to be redone as a PR for the mainline [openjdk/jdk](https://github.com/openjdk/jdk) repo.

PR #257, opened 2021-07-20
8270085: Suspend during block transition may deadlock if lock held

pchilano:
Hi all,

The following patch fixes deadlocks issues that could occur when checking for suspension while holding VM locks. See the bug description for a concrete case. The solution is to avoid checking for suspend requests when using the TBIVM wrapper. The original patch was actually written by @robehn(now on vacations) and I only made small changes.

Tested in mach5 tiers1-7. I also verified that the new added test SuspendBlocked.java deadlocks with the current bits and passes with this patch.

Thanks,
 Patricio

---------
### Progress

### Issue
 * [JDK-8270085](https://bugs.openjdk.java.net/browse/JDK-8270085): Suspend during block transition may deadlock if lock held

### Reviewers
 * [Daniel D. Daugherty](https://openjdk.java.net/census#dcubed) (@dcubed-ojdk - **Reviewer**)
 * [Coleen Phillimore](https://openjdk.java.net/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to 6ad7c4ea32925e515d9bf228276a2794135d4897

### Contributors
 * Robbin Ehn `<rehn@openjdk.org>`
 * Patricio Chilano Mateo `<pchilanomate@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/257/head:pull/257` \
`$ git checkout pull/257`

`$ git checkout pull/257` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/257/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 257`

`$ git pr show -t 257`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/257.diff">https://git.openjdk.java.net/jdk17/pull/257.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/handshake.cpp
+++ b/src/hotspot/share/runtime/handshake.cpp
@@ -77,6 +77,7 @@ class HandshakeOperation : public CHeapObj<mtThread> {
   int32_t pending_threads()        { return Atomic::load(&_pending_threads); }
   const char* name()               { return _handshake_cl->name(); }
   bool is_async()                  { return _handshake_cl->is_async(); }
+  bool is_suspend()                { return _handshake_cl->is_suspend(); }
 };
 
 class AsyncHandshakeOperation : public HandshakeOperation {
@@ -376,6 +377,7 @@ void Handshake::execute(HandshakeClosure* hs_cl, JavaThread* target) {
     // Check for pending handshakes to avoid possible deadlocks where our
     // target is trying to handshake us.
     if (SafepointMechanism::should_process(self)) {
+      // Will not suspend here.
       ThreadBlockInVM tbivm(self);
     }
     hsy.process();
@@ -425,11 +427,19 @@ bool HandshakeState::operation_pending(HandshakeOperation* op) {
   return _queue.contains(mo);
 }
 
-HandshakeOperation* HandshakeState::get_op_for_self() {
+static bool no_suspend_filter(HandshakeOperation* op) {
+  return !op->is_suspend();
+}
+
+HandshakeOperation* HandshakeState::get_op_for_self(bool allow_suspend) {
   assert(_handshakee == Thread::current(), "Must be called by self");
   assert(_lock.owned_by_self(), "Lock must be held");
-  return _queue.peek();
-};
+  if (allow_suspend) {
+    return _queue.peek();
+  } else {
+    return _queue.peek(no_suspend_filter);
+  }
+}
 
 static bool non_self_queue_filter(HandshakeOperation* op) {
   return !op->is_async();
@@ -439,6 +449,11 @@ bool HandshakeState::have_non_self_executable_operation() {
   assert(_handshakee != Thread::current(), "Must not be called by self");
   assert(_lock.owned_by_self(), "Lock must be held");
   return _queue.contains(non_self_queue_filter);
+}
+
+bool HandshakeState::has_a_non_suspend_operation() {
+  MutexLocker ml(&_lock, Mutex::_no_safepoint_check_flag);
+  return _queue.contains(no_suspend_filter);
 }
 
 HandshakeOperation* HandshakeState::get_op() {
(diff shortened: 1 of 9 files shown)


dcubed-ojdk (approved):
Thumbs up.

pchilano:
> Thumbs up.
>
Thanks for the review Dan!

pchilano:
Closing and using the jdk17 PR: https://github.com/openjdk/jdk17/pull/257 instead.

dholmes-ora (approved):
Looks good.

Sorry I missed the fact this should have been based on 17 to begin with.

src/hotspot/share/runtime/handshake.cpp
```diff
@@ -454,25 +469,22 @@ void HandshakeState::remove_op(HandshakeOperation* op) {
...
-    // The exception to this rule is the asynchronous suspension handshake.
-    // It by-passes the NSV by manually doing the transition.
-    NoSafepointVerifier nsv;
-    return process_self_inner();
-  }
-}
+  // Handshakes cannot safely safepoint.
+  // The exception to this rule is the asynchronous suspension handshake.
+  // It by-passes the NSV by manually doing the transition.
+  NoSafepointVerifier nsv;
```
coleenp, line 482: I have no idea what this comment means in this context. Below we take out a lock with _no_safepoint_check which is essentially a NSV. You just moved this comment so I don't suggest changing it at this time.

src/hotspot/share/runtime/handshake.cpp
```diff
@@ -667,8 +680,15 @@ class SuspendThreadHandshake : public HandshakeClosure {
...
 bool HandshakeState::suspend() {
+  JavaThread* self = JavaThread::current();
   SuspendThreadHandshake st;
   Handshake::execute(&st, _handshakee);
+  if (_handshakee == self) {
+    // If target is the current thread we need to call this to do the
+    // actual suspend since Handshake::execute() above only installed
+    // the asynchronous handshake.
+    SafepointMechanism::process_if_requested(self);
+  }
```
coleenp, line 691: Is this an optimization? Or can the thread escape the suspend request?

test/hotspot/jtreg/runtime/handshake/SuspendBlocked.java
```diff
@@ -0,0 +1,58 @@
...
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+/*
+ * @test SuspendBlocked
+ * @bug 8270085
+ * @library /testlibrary /test/lib
```
coleenp, line 28: I think this is referring to /testlibrary that I removed. Can you see if it's still needed?

coleenp (approved):
Just some comments and questions. Looks good!

pchilano:
> Looks good.
> 
> Sorry I missed the fact this should have been based on 17 to begin with.
>
Thanks for the review David!

src/hotspot/share/runtime/handshake.cpp
pchilano: Right, it's created with _allow_vm_block = true so the NSV is redundant. The _no_safepoint_check actually means we don't check for safepoints if there is contention while acquiring the monitor but we don't do inc_no_safepoint_count() to enforce no safepoints later though.

src/hotspot/share/runtime/handshake.cpp
pchilano: For self-suspend requests coming from Java we will suspend when going back to Java. In the case of JVMTI self-suspend requests I think Robbin found some JVMTI tests that were expecting to not return back to native without suspending.

test/hotspot/jtreg/runtime/handshake/SuspendBlocked.java
pchilano: Not needed, removed.

pchilano:
> Just some comments and questions. Looks good!
>
Thanks for the review Coleen!

src/hotspot/share/runtime/handshake.cpp
coleenp: A side effect of the no_safepoint_check locking is that it has an implicit NSV though.  We do call inc_no_safepoint_count when setting the owner, but I remembered this wrong. It's not unconditional:

    // NSV implied with locking allow_vm_block flag.
    // The tty_lock is special because it is released for the safepoint by
    // the safepoint mechanism.
    if (new_owner->is_Java_thread() && _allow_vm_block && this != tty_lock) {
      JavaThread::cast(new_owner)->inc_no_safepoint_count();
    }

src/hotspot/share/runtime/handshake.cpp
coleenp: We should document the order of events why this code is needed in a future RFE, because this comment doesn't really explain it to me and knowing where to put all these process_if_requested() and process() calls is far from obvious.  I believe you and Robbin, but there are too many mysteries.

coleenp (approved):
Still good.

dcubed-ojdk (approved):
Still good.

src/hotspot/share/runtime/handshake.cpp
pchilano: I commented the line and run all the serviceability/ tests to look for issues. The only test that failed was SuspendWithCurrentThread.java due to the recent 8264663 change (https://github.com/openjdk/jdk/commit/b5c6351315eeec64fe9c06fc7c8719efa08dc8a6). The patch aimed to verify that the self-suspending thread actually suspended and didn't exited, but the check is too restrictive because it doesn't consider that the JT will still suspend when going back to Java. That check can be moved to Java to verify the suspender didn't return from suspendTestedThreads().
Maybe I should do that in another RFE along with removing this code?

pchilano:
While chatting with Coleen I realized the comment in should_process_no_suspend() wasn't accurate so I fixed it.  Not necessarily there will be a suspend request when reaching that path. The poll could be armed just because a safepoint or a non-suspend handshake operation was executed while the thread was safepoint safe.

dcubed-ojdk (approved):
Thanks for adjusting the comment. Still good.

src/hotspot/share/runtime/handshake.cpp
coleenp: Yes, that would be good to clear up in a future RFE.

dholmes-ora (approved):

pchilano:
Thanks @dholmes-ora, @dcubed-ojdk and @coleenp for the reviews!

PR #283, opened 2021-07-26
8271251: JavaThread::java_suspend() fails with "fatal error: Illegal threadstate encountered: 6"

pchilano:
Hi,

Please review this small patch. When self-suspending, a JavaThread might reach SafepointSynchronize::block() with a state of _thread_in_vm which is not listed as a valid state for safepoint polling. There are a couple of simple ways to fix this. As suggested by @dholmes-ora  we can avoid the handshake machinery altogether and directly self-suspend. Since this issue is intermittent and in tier7 which I didn't run as many times as other lower tiers it escaped my testing of 8270085.

Testing in mach5 tiers1-7. I also reproduced the test failures locally and verified that now both hs202t002.java and ThreadSuspendSelf.java are passing. 

Thanks,
Patricio

---------
### Progress

### Issue
 * [JDK-8271251](https://bugs.openjdk.java.net/browse/JDK-8271251): JavaThread::java_suspend() fails with "fatal error: Illegal threadstate encountered: 6"

### Reviewers
 * [Daniel D. Daugherty](https://openjdk.java.net/census#dcubed) (@dcubed-ojdk - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/283/head:pull/283` \
`$ git checkout pull/283`

`$ git checkout pull/283` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/283/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 283`

`$ git pr show -t 283`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/283.diff">https://git.openjdk.java.net/jdk17/pull/283.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/handshake.cpp
+++ b/src/hotspot/share/runtime/handshake.cpp
@@ -612,6 +612,7 @@ void HandshakeState::do_self_suspend() {
   assert(Thread::current() == _handshakee, "should call from _handshakee");
   assert(_lock.owned_by_self(), "Lock must be held");
   assert(!_handshakee->has_last_Java_frame() || _handshakee->frame_anchor()->walkable(), "should have walkable stack");
+  assert(_handshakee->thread_state() == _thread_blocked, "Caller should have transitioned to _thread_blocked");
 
   log_trace(thread, suspend)("JavaThread:" INTPTR_FORMAT " suspended", p2i(_handshakee));
   while (is_suspended()) {
--- a/src/hotspot/share/runtime/handshake.cpp
+++ b/src/hotspot/share/runtime/handshake.cpp
@@ -614,8 +614,8 @@ void HandshakeState::do_self_suspend() {
   assert(!_handshakee->has_last_Java_frame() || _handshakee->frame_anchor()->walkable(), "should have walkable stack");
   assert(_handshakee->thread_state() == _thread_blocked, "Caller should have transitioned to _thread_blocked");
 
-  log_trace(thread, suspend)("JavaThread:" INTPTR_FORMAT " suspended", p2i(_handshakee));
   while (is_suspended()) {
+    log_trace(thread, suspend)("JavaThread:" INTPTR_FORMAT " suspended", p2i(_handshakee));
     _lock.wait_without_safepoint_check();
   }
   log_trace(thread, suspend)("JavaThread:" INTPTR_FORMAT " resumed", p2i(_handshakee));
--- a/src/hotspot/share/runtime/handshake.cpp
+++ b/src/hotspot/share/runtime/handshake.cpp
@@ -612,14 +612,11 @@ void HandshakeState::do_self_suspend() {
   assert(Thread::current() == _handshakee, "should call from _handshakee");
   assert(_lock.owned_by_self(), "Lock must be held");
   assert(!_handshakee->has_last_Java_frame() || _handshakee->frame_anchor()->walkable(), "should have walkable stack");
-  JavaThreadState jts = _handshakee->thread_state();
+
+  log_trace(thread, suspend)("JavaThread:" INTPTR_FORMAT " suspended", p2i(_handshakee));
   while (is_suspended()) {
-    _handshakee->set_thread_state(_thread_blocked);
-    log_trace(thread, suspend)("JavaThread:" INTPTR_FORMAT " suspended", p2i(_handshakee));
     _lock.wait_without_safepoint_check();
   }
-  _handshakee->set_thread_state(jts);
-  set_async_suspend_handshake(false);
   log_trace(thread, suspend)("JavaThread:" INTPTR_FORMAT " resumed", p2i(_handshakee));
 }
 
@@ -631,7 +628,12 @@ class ThreadSelfSuspensionHandshake : public AsyncHandshakeClosure {
   void do_thread(Thread* thr) {
     JavaThread* current = thr->as_Java_thread();
     assert(current == Thread::current(), "Must be self executed.");
+    JavaThreadState jts = current->thread_state();
+
+    current->set_thread_state(_thread_blocked);
     current->handshake_state()->do_self_suspend();
+    current->set_thread_state(jts);
+    current->handshake_state()->set_async_suspend_handshake(false);
   }
   virtual bool is_suspend() { return true; }
 };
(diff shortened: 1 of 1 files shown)


src/hotspot/share/runtime/handshake.cpp
```diff
@@ -612,14 +612,11 @@ void HandshakeState::do_self_suspend() {
   assert(Thread::current() == _handshakee, "should call from _handshakee");
   assert(_lock.owned_by_self(), "Lock must be held");
   assert(!_handshakee->has_last_Java_frame() || _handshakee->frame_anchor()->walkable(), "should have walkable stack");
-  JavaThreadState jts = _handshakee->thread_state();
+
```
dholmes-ora, line 615: Can we add an assert:

`assert(_handshakee->thread_state() == _thread_blocked, "Caller should have transitioned to _thread_blocked);
`
?

src/hotspot/share/runtime/handshake.cpp
```diff
@@ -612,14 +612,11 @@ void HandshakeState::do_self_suspend() {
...
+  log_trace(thread, suspend)("JavaThread:" INTPTR_FORMAT " suspended", p2i(_handshakee));
```
dholmes-ora, line 616: I think I prefer this logging to remain in the while loop. By placing it there you know the thread will actually wait() and you can also see if there are multiple suspend requests. It might even make sense to put the resume log statement in the loop after the wait() with an explicit check of is_suspended() to preclude a spurious wakeup.

dholmes-ora (approved):
Looks good! Thanks for fixing this Patricio.

A couple of minor suggestions.

David

src/hotspot/share/runtime/handshake.cpp
pchilano: Done!

src/hotspot/share/runtime/handshake.cpp
pchilano: So for the resume log statement, once the target wakes up from wait() the _suspended flag could be set not necessarily because of a spurious wakeup but because another JT suspended the target right after the resume. But I can add a log statement in HandshakeState::resume() to track all resumes. 
As for the suspend log statement I thought about not moving it but then I saw that we are already logging suspends in HandshakeState::suspend_with_handshake(), and since all these logging statements are done under _lock held there is already a sequence of suspend/resume events (if adding log in HandshakeState::resume()). 
What do you think?

src/hotspot/share/runtime/handshake.cpp
dcubed-ojdk: Should be able to add that assert. The new code has only two callers
to do_self_suspend() and both make sure we are in `_thread_blocked`.

src/hotspot/share/runtime/handshake.cpp
dcubed-ojdk: I would not change any of the logging in the JDK17 fix unless there's a very
good reason. We want the JDK17 fix to be as minimal as possible.

pchilano: Sounds good, I added the log statement back into the loop.

dcubed-ojdk:
@pchilano - I've reviewed v0 and v2. I want to mull on this over night so
I'll post my final review in the AM.

dholmes-ora (approved):
Updates are good.

Thanks,
David

pchilano:
> Updates are good.
> 
Thanks for the review David!

Patricio

dcubed-ojdk:
In `src/hotspot/share/runtime/safepointMechanism.cpp`, I would love to see:

```
void SafepointMechanism::process(JavaThread *thread, bool allow_suspend) {
  // Read global poll and has_handshake after local poll
  OrderAccess::loadload();

  // local poll already checked, if used.
  bool need_rechecking;
  do {
    assert(is_a_block_safe_state(thread->thread_state()), "thread state is not safe for a SafepointSynchronize::block()");
    if (global_poll()) {
      // Any load in ::block() must not pass the global poll load.
      // Otherwise we might load an old safepoint counter (for example).
      OrderAccess::loadload();
      SafepointSynchronize::block(thread);
    }
```

We don't have an `is_a_block_safe_state()` function so I would be okay with
a simple check like this:
```
  assert(thread->thread_state() != _thread_in_vm,  "thread state is not safe for a SafepointSynchronize::block()");
```

I looked at both `SafepointSynchronize::handshake_safe()` and `safepoint_safe_with()`
and neither of those is quite right.

dcubed-ojdk (approved):

dcubed-ojdk:
I'm okay with the changes as they are, but I'd love to see an assert()
added in `SafepointMechanism::process()` as I describe above. I think
this would help catch any other unexpected go-to-a-safepoint from
the wrong state errors.

dcubed-ojdk:
If you decide to add `is_a_block_safe_state()`, then add it above
SafepointSynchronize::block() and use it to determine what safepoint
safe states are accepted by SafepointSynchronize::block(), i.e., share
the new function.

pchilano:
> If you decide to add `is_a_block_safe_state()`, then add it above
> SafepointSynchronize::block() and use it to determine what safepoint
> safe states are accepted by SafepointSynchronize::block(), i.e., share
> the new function.
>
So having an assert only for the _thread_in_vm case doesn't seem right to me since it would test only for this specific case that we are fixing. I can add the is_a_block_safe_state() method, but if we want to avoid code and check duplication I would have to remove the current switch statement from SafepointSynchronize::block(). Then the assert you want can be made a guarantee. Something like this:

https://github.com/pchilano/jdk17/commit/db0f8eb4dba08d6a7250bd9e87816f2f024fa600

I defined is_a_block_safe_state() in safepoint.hpp to allow inlining in SafepointMechanism::process().

Unless you want to keep the check of state in ::block, in which case I can add the guarantee in ::block instead and leave it in process() as an assert. The duplicate check would be only in debug mode in that case.

vnkozlov (commented):
Note, currently you need to get approval for JDK 17 fix according to: 
http://openjdk.java.net/jeps/3#Fix-Request-Process

pchilano:
Hi Vladimir,

> Note, currently you need to get approval for JDK 17 fix according to:
> http://openjdk.java.net/jeps/3#Fix-Request-Process
>
Right, I was waiting for the 2 reviews before posting the fix request. I'll do that right after all comments are resolved.

Thanks,
Patricio

pchilano:
> If you decide to add `is_a_block_safe_state()`, then add it above
> SafepointSynchronize::block() and use it to determine what safepoint
> safe states are accepted by SafepointSynchronize::block(), i.e., share
> the new function.
>
So as we discussed offline in order to minimize changes going to 17 and given that it's simple to figure out by code inspection that all callers of process_if_requested() are either in a trans or Java state and the issue with the wrong state in 8270085 is clearly fixed, I created 8271348 to add the sanity check in 18 to guard against future code changes falling in this same issue.

pchilano:
Thanks for the review Dan!

dcubed-ojdk:
Thanks for filing:

JDK-8271348 Add stronger sanity check of thread state when polling for safepoint/handshakes
https://bugs.openjdk.java.net/browse/JDK-8271348

Still thumbs up!

PR #290, opened 2021-07-28
8067223: [TESTBUG] Rename Whitebox API package

iignatev:
Hi all,

could you please review this big tedious and trivial(-ish) patch which moves `sun.hotspot.WhiteBox` and related classes to `jdk.test.whitebox` package?

the majority of the patch is the following substitutions:
 - `s~sun/hotspot/WhiteBox~jdk/test/whitebox/WhiteBox~g`
 - `s/sun\.hotspot\.parser/jdk.test.whitebox.parser/g`
 - `s/sun\.hotspot\.cpuinfo/jdk.test.whitebox.cpuinfo/g`
 - `s/sun\.hotspot\.code/jdk.test.whitebox.code/g`
 - `s/sun\.hotspot\.gc/jdk.test.whitebox.gc/g`
 - `s/sun\.hotspot\.WhiteBox/jdk.test.whitebox.WhiteBox/g`

testing: tier1-4

Thanks,
-- Igor

---------
### Progress

### Issue
 * [JDK-8067223](https://bugs.openjdk.java.net/browse/JDK-8067223): [TESTBUG] Rename Whitebox API package

### Reviewers
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to 237e886049b652b8eb0dad6991f52e749671ec13
 * [Vladimir Kozlov](https://openjdk.java.net/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to 237e886049b652b8eb0dad6991f52e749671ec13

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk17 pull/290/head:pull/290` \
`$ git checkout pull/290`

`$ git checkout pull/290` \
`$ git pull https://git.openjdk.java.net/jdk17 pull/290/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 290`

`$ git pr show -t 290`

</details>

<a href="https://git.openjdk.java.net/jdk17/pull/290.diff">https://git.openjdk.java.net/jdk17/pull/290.diff</a>

</details>


vnkozlov:
I know that tests fixes could be pushed during RDP2 without approval.
But these one is very big and it is not a fix - it is enhancement.

What is the reason you want to push it into JDK 17 just few days before first Release Candidate? Instead of pushing it into JDK 18.

And I can't even review it because GutHub UI hangs on these big changes.

dholmes-ora:
I agree with @vnkozlov this too big and disruptive for 17 at this stage of the release.

I also think a better approach to this cleanup would have been to copy the WhiteBox class to the new package structure, then update the tests in chunks, then remove the old WhiteBox classes when that is complete.

Thanks,
David

iignatev:
@vnkozlov, @dholmes-ora,

Thank you for looking at this!

I want this to be integrated into JDK 17 b/c some "external" libraries use (used to use) WhiteBox API, e.g. jcstress[[2]] used WhiteBox API to deoptimize compiled methods[[3]], and it would be easier for maintainers of such libraries to condition package name based on JDK major version. Also, given JDK 17 is an LTS, it would be beneficial for everyone not to have big differences in test bases b/w it and the mainline.

according to JEP 3, test RFEs are allowed until the very end and don't require late enhancement approval: "Enhancements to tests and documentation during RDP 1 and RDP 2 do not require approval, as long as the relevant issues are identified with a noreg-self or noreg-doc label, as appropriate"[[1]]. So, process-wise, I don't see any issues w/ integrating this RFE, yet, I fully agree that due to its size, this patch can be disruptive and can cause massive failures, which is something we obviously don't want at the current stage of JDK 17. 

I like David's idea about phasing this clean-up, and, due to the reasons described above, I would like to integrate the first part, copying WhiteBox classes to the new package structure and associated changes w/o updating all the tests, into JDK 17 and update the tests on the mainline (w/ backporting into jdk17u).

WDYT?

Cheers,
-- Igor

[1]: https://openjdk.java.net/jeps/3#Late-Enhancement-Request-Process
[2]: https://github.com/openjdk/jcstress
[3]: https://github.com/openjdk/jcstress/blob/df83b446f187ae0b0fa31fa54decb59db9e955da/jcstress-core/src/main/java/org/openjdk/jcstress/vm/WhiteBoxSupport.java

iignatev:
Vladimir, David,

I've (forced) pushed a smaller version of the renaming. instead of removing `sun.hotspot` classes, it copies them to `jdk.test.whitebox` (w/ `s.h.parser.DiagnosticCommand` being removed as it's used in WhiteBox signature and it was easier to update a few tests that use it), updates hotspot code to register native methods for both `sun.hotspot.WhiteBox` and `jdk.test.whitebox.WhiteBox` classes. To make it easier and not to introduce extra dependency, I've made it impossible to use `s.h.WB` w/ a security manager enabled, otherwise there would be a dependency b/w `s.h.WB` and `j.t.w.WB$WhiteBoxPermission` or there would be 2 permissions. There are no open JDK tests that are impacted by this limitation.

With minor tweaks in closed source, the patch successfully passes Oracle's tier1-4.

-- Igor

dholmes-ora (approved):
Hi Igor,

This set of changes seems much more manageable to me.

Not sure about the new deprecation warnings for the old WB classes .. might that not itself trigger some failures? If not then I don't see how the deprecation warnings help with transitioning to the new WB class?

Thanks,
David

DamonFool:
Shall we also update the copyright year like test/lib/sun/hotspot/cpuinfo/CPUInfo.java?
Thanks.

vnkozlov (approved):
I agree with these revised changes for JDK 17.

iignatev:
Hi David,

> This set of changes seems much more manageable to me.

thank you for your review, David.

> Not sure about the new deprecation warnings for the old WB classes .. might that not itself trigger some failures? If not then I don't see how the deprecation warnings help with transitioning to the new WB class?

the deprecation warnings (hopefully) will help people not to forget that they should use the new WB class in new tests. 

Thanks,
-- Igor

iignatev:
Hi Jie,
> Shall we also update the copyright year like test/lib/sun/hotspot/cpuinfo/CPUInfo.java?

we most certainly shall, just pushed the commit that updates the copyright years in the touched files.

Cheers,
-- Igor

iignatev:
> I agree with these revised changes for JDK 17.

Thanks for your review, Vladimir. 

I'll rerun my testing before integrating (just for good luck).

-- Igor

